PR #1, opened 2018-10-16
Allow use as a meson subproject

jtojnar:


meson.build
```diff
@@ -89,12 +89,12 @@ test_link_args = [
   '-Wl,-z,relro',
   '-Wl,-z,now',
 ]
-foreach arg: test_link_args
-  if cc.has_link_argument(arg)
-    global_link_args += arg
+foreach link_arg: test_link_args
+  if cc.links('int main () { return 0; }', name: link_arg, args: link_arg)
```
hughsie: is this a workaround? Seems a little hacky...

jtojnar: You are right, this is a hack, I did not notice has_link_argument was added in 0.45.

hughsie:
@jtojnar the self tests are using non-exported functions, hence why I was building them "again" as part of the  xb-self-test binary.

hughsie:
I've tweaked this, and merged it. I do get some pretty alarming warnings however:

    src/libxmlb/meson.build:5: WARNING: Got an empty configuration_data() object and found no substitutions in the input file 'xb-builder.h'. If you want to copy a file to the build dir, use the 'copy:' keyword argument added in 0.47.0
    WARNING: Output file "src/libxmlb/@PLAINNAME@" for configure_file() at src/libxmlb/meson.build:5 overwrites configure_file() output at src/libxmlb/meson.build:5
    Configuring xb-builder-node.h using configuration
    src/libxmlb/meson.build:5: WARNING: Got an empty configuration_data() object and found no substitutions in the input file 'xb-builder-node.h'. If you want to copy a file to the build dir, use the 'copy:' keyword argument added in 0.47.0
    WARNING: Output file "src/libxmlb/@PLAINNAME@" for configure_file() at src/libxmlb/meson.build:5 overwrites configure_file() output at src/libxmlb/meson.build:5
    Configuring xb-builder-source.h using configuration
    src/libxmlb/meson.build:5: WARNING: Got an empty configuration_data() object and found no substitutions in the input file 'xb-builder-source.h'. If you want to copy a file to the build dir, use the 'copy:' keyword argument added in 0.47.0

jtojnar:
Yes, that is expected. We can change it to use copy kwarg once we upgrade to meson 0.47.

hughsie:
@jtojnar  -- I think we can do that now; Centos even has 0.47.2 -- can you do a PR with that change please? Thanks!

jtojnar:
I pushed one more fix to the subproject branch. And opened PR about the `copy` kwarg #2.

PR #2, opened 2018-10-17
meson: Copy headers using copy kwarg

jtojnar:
Meson 0.47 introduces copy kwarg [1] to configure_file, allowing us to copy files to build directory in a slightly cleaner way than passing an empty configuration_data() object.

It is still a hack before install_headers can copy files to subdir in the build directory. [2]

[1]: http://mesonbuild.com/Reference-manual.html#configure_file
[2]: https://github.com/mesonbuild/meson/issues/2546

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('libxmlb', 'c',
   version : '0.1.3',
   license : 'LGPL-2.1+',
-  meson_version : '>=0.46.0',
+  meson_version : '>=0.47.0',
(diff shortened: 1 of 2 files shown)


hughsie:
I wonder, would it solve the issue if we moved the files from `src` to `libxmlb` directories?

jtojnar:
It would but source code in the `src` is more elegant. And since this is only a temporary measure, I would prefer this hack to moving the files back and forth.

PR #5, opened 2018-10-28
Don't hardcode the path of python3

lantw44:
Most *BSD systems don't consider python as an integral part of the
operating system, so python is seldom installed in /usr prefix.

This fixes gnome-software build on FreeBSD:
```
[38/259] Generating libxmlb-mapfile with a custom command.
FAILED: subprojects/libxmlb/src/libxmlb.map
/home/lantw44/gnome/source/gnome-software/subprojects/libxmlb/src/generate-version-script.py LIBXMLB subprojects/libxmlb/src/Xmlb-1.0.gir subprojects/libxmlb/src/libxmlb.map
/bin/sh: /home/lantw44/gnome/source/gnome-software/subprojects/libxmlb/src/generate-version-script.py: not found
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -123,6 +123,9 @@ configure_file(
   configuration : conf
 )
 
+python = import('python')
+python3 = python.find_installation('python3')
+
 subdir('src')
 if get_option('gtkdoc')
   gtkdocscan = find_program('gtkdoc-scan', required : true)
--- a/src/generate-version-script.py
+++ b/src/generate-version-script.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 # pylint: disable=invalid-name,missing-docstring
 #
 # Copyright (C) 2017 Richard Hughes <richard@hughsie.com>
--- a/src/meson.build
+++ b/src/meson.build
@@ -162,6 +162,7 @@ if get_option('introspection')
     input: gir[0],
     output: 'libxmlb.map',
     command: [
+      python3,
       join_paths(meson.current_source_dir(), 'generate-version-script.py'),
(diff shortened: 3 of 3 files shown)


hughsie:
Isn't http://julio.meroh.net/2016/09/env-considered-harmful.html still valid? I guess we could make it a meson argument, but I don't think adding back `env` is the right way to fix this.

superm1:
I think the way it was solved in fwupd is sufficient to apply here too no?
https://github.com/hughsie/fwupd/commit/669d4382a071aa5977efc03d9a8fd14cb8ab1e72#diff-969b60ad3d206fd45c208e266ccfed38

hughsie:
That works for me certainly... @lantw44 -- would you be able to prepare a patch doing the same thing for libxmlb too? Thanks.

lantw44:
> Isn't http://julio.meroh.net/2016/09/env-considered-harmful.html still valid? I guess we could make it a meson argument, but I don't think adding back `env` is the right way to fix this.

I agree that `#!/usr/bin/env` should not be used in packaged or installed files, and it is better for the build system to find required interpreters instead of relying on shebangs. However, I think `#!/usr/bin/env` is usually good enough for uninstalled scripts because it works most of the time and most build commands already depend on `PATH` to work properly. If using `#!/usr/bin/env` isn't allowed, anything other than `#!/bin/sh` is likely to stop working when scripts are moved to different systems.

> That works for me certainly... @lantw44 -- would you be able to prepare a patch doing the same thing for libxmlb too? Thanks.

I will upload a patch to use `python.find_installation('python3')` to find python3.

PR #9, opened 2019-01-28
libxmlb: Treat text/plain as MIME for xml

djcampello:
.xml files in ChromeOS do not have the MIME type expected by the
library. Adding the MIME type that is currently reported on the files by
the OS in the list of MIME types handled by the library.


hughsie:
Why don't the XML files get the right content type detected? I think this will cause other problems further down the line if we don't fix the root cause.

hughsie:
My view here is that `text/plain` just isn't right. But, I'd be fine adding some heuristics to `xb_builder_source_ctx_get_content_type()` if content-type detection can't be enabled for some reason.

PR #11, opened 2019-02-08
Fix self tests on port architectures

superm1:
This hasn't actually been tested on the buildds yet (like I said I need to commit a patch into a build), so I wanted to make sure it was acceptable first.


hughsie:
Gahh, I committed basically that a few hours ago, sorry!

superm1:
Ah okay 🥙

Issue #13, opened 2019-02-22
Converters don't nest - can't handle .yml.gz


robert-ancell:
libxmlb assumes the result of a gzipped file is XML, instead of checking the MIME type of the decompressed data. This means that .yml.gz files can't be handled by the converter.

hughsie:
Got a patch handy? I'm just on PTO. Thanks.

robert-ancell:
I will attempt to write one but haven't started yet. I figure the best thing is to re-check the mime type after a gzip decode.

robert-ancell:
https://github.com/hughsie/libxmlb/pull/14

hughsie:
Okay, I've had a go at this too; I took what you did and rewrote it -- there is a proper context object we can extend in the future if we need to now. I've also pushed a wip/hughsie/XbBuilderSource-adapter branch in gnome-software which uses this new API -- feedback very welcome.

PR #14, opened 2019-02-24
Convert the contents after decoding a GZipped file

robert-ancell:
Previously the contents were assumed to be XML, and this didn't work with GZip
compressed YAML.

The code isn't optimal as it writes decodes the GZipped file to a temporary
file. This was done because the converter API uses a GFile instead of a
GInputStream. The existing XML case doesn't use the temporary file so shouldn't
have a performance hit.

### DIFF
--- a/src/xb-builder-source.c
+++ b/src/xb-builder-source.c
@@ -448,19 +448,94 @@ xb_builder_source_load_plain_cb (XbBuilderSource *self,
 }
 
 static GInputStream *
+get_gzip_stream (GFile *file,
+		 GCancellable *cancellable,
+		 GError **error)
+{
+	g_autoptr(GInputStream) istream = NULL;
+	g_autoptr(GZlibDecompressor) decompressor = NULL;
+
+	istream = G_INPUT_STREAM (g_file_read (file, cancellable, error));
+	if (istream == NULL)
+		return NULL;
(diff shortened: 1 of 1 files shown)


robert-ancell:
https://github.com/hughsie/libxmlb/issues/13

hughsie:
I tried to implement something a bit cleaner (i.e. using a GInputStream for the vfunc) but I couldn't get `g_file_input_stream_query_info()` to return the `G_FILE_ATTRIBUTE_STANDARD_CONTENT_TYPE` -- only NULL for a valid stream of `<id>gimp</id>`.  Any tips? I tried seeking to the file start before checking but nothing.

robert-ancell:
I also tried using `g_file_input_stream_query_info()` initially but looking at the glib source only `G_FILE_ATTRIBUTE_TIME_MODIFIED`, `G_FILE_ATTRIBUTE_TIME_MODIFIED_USEC`, and `G_FILE_ATTRIBUTE_STANDARD_NAME` are set (`gio/gzlibdecompressor.c` `g_zlib_decompressor_convert()`). So that's why I had to use `g_content_type_guess()`.

Note if you change the API to use `GInputStream` then you also need to pass the filename around. In the case of YAML the `g_content_type_guess()` doesn't work without checking for the extension.

superm1:
Superseded by #15.  Feel free to leave any comments on it there.

PR #15, opened 2019-02-27
Allow nesting XbBuilderSource content type handlers

hughsie:
This is much more flexible than the existing GFile -> GInputStream converter
functions. Content types can be "nested" for instance:

    application/gzip -> application/x-yaml -> application/xml

To ensure the content type is always correct when there is no filename info,
always add the XML header when building sources.

For API compatibility we also have to support the deprecated vfunc too.

Fixes https://github.com/hughsie/libxmlb/issues/13


src/xb-builder-source-ctx.c
```diff
@@ -0,0 +1,165 @@
...
+	}
+	return g_content_type_guess (priv->filename, buf, bufsz, NULL);
+}
+
+/* private */
+void
+xb_builder_source_ctx_set_filename (XbBuilderSourceCtx *self, const gchar *filename)
+{
+	XbBuilderSourceCtxPrivate *priv = GET_PRIVATE (self);
+	g_return_if_fail (XB_IS_BUILDER_SOURCE_CTX (self));
```
superm1: I think you'll want to also fail if filename is NULL too right?

src/xb-builder-source.c
```diff
@@ -298,6 +314,10 @@ xb_builder_source_add_node_func (XbBuilderSource *self,
  * Adds a function that can be used to convert files loaded with
  * xb_builder_source_load_xml().
  *
+ * This function is now deprecated, and xb_builder_source_add_adapter() can be
+ * used instead. The latter allows chaining streams rather than just one #GFile
+ * to #GInputStream conversion.
```
superm1: Maybe worth doing a g_warning or similar that this is a deprecated function to discourage use

src/xb-builder-source.c
```diff
@@ -394,31 +453,77 @@ xb_builder_source_get_info (XbBuilderSource *self)
...
+		g_autoptr(GInputStream) istream_tmp = NULL;
+
+		/* get the content type of the stream */
+		xb_builder_source_ctx_set_filename (ctx, basename);
+		content_type = xb_builder_source_ctx_get_content_type (ctx,
+								       cancellable,
+								       error);
+		if (content_type == NULL)
+			return NULL;
+		if (g_strcmp0 (content_type, "application/xml") == 0)
```
superm1: Considering #9, is looking for text/plain another alternative?  Or should #9 be rejected outright and don't support unless proper mime types are used?

superm1:
Considering this is actively going to affect gnome-software from those bug reports mentioned it's probably worth spinning a release fairly soon after merging.

src/xb-builder-source-ctx.c
hughsie: Agree, thanks.

src/xb-builder-source.c
hughsie: `text/plain` isn't right. I think #9 should be closed -- I'd be fine adding some heuristics to `xb_builder_source_ctx_get_content_type()` if content-type detection can't be enabled for some reason.

src/xb-builder-source.c
hughsie: I think build time it's useful, but not so much runtime. When you build an old gnome-software with this branch you get:

    ../plugins/flatpak/gs-flatpak.c: In function ‘gs_flatpak_load_desktop_fn’:
    ../plugins/flatpak/gs-flatpak.c:585:2: warning: ‘xb_builder_source_add_converter’ is deprecated: Use 'xb_builder_source_add_adapter' instead [-Wdeprecated-declarations]

robert-ancell:
This might also need an AppStream glib change to make an AsStore object from and input stream - it currently requires a GFile or GBytes (the latter won't detect it's YAML).

hughsie:
> the latter won't detect it's YAM

Urgh, I didn't think of that. Is there any way to mark the YAML file so that is can be recognized by g_content_type_guess()? If not, I wouldn't be against adding manual matching of

    ---
    File: DEP-11

in `as_store_from_bytes()`...

robert-ancell:
I think because YAML is so super-generic it would be pretty hard to make the MIME parser detect it easily (note I'm not a YAML expert by any means). I think that adding the DEP-11 check is a good idea  onto appstream-glib - because in general YAML won't work and all DEP-11 file are required to have that field. So it seems a reliable method of detecting content.

PR #17, opened 2019-03-07
Add XB_BUILDER_COMPILE_FLAG_IGNORE_GUID

hughsie:
This allows us to load from a silo without caring about the exact GUID.
This may be useful if the cache keys are not the same for some reason.


hughsie:
We'll be using this from fwupd, so that we can run in fwupd.shutdown without accidentally regenerating a cache on a read only filesystem.

PR #18, opened 2019-03-12
Allow compilation on older versions of meson

superm1:
There is a general agreement to bring fwupd 1.2.x into Ubuntu 18.04, so I'm sorting out some of the changes needed for this.

This reverts commit 96437763b2b179e402e2660c06a0036397c29fcd and makes adjustments to dynamically decide what to do based upon meson version.

This is intended to be used for backporting libxmlb to 18.04.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('libxmlb', 'c',
   version : '0.1.8',
   license : 'LGPL-2.1+',
-  meson_version : '>=0.47.0',
+  meson_version : '>=0.45.0',
   default_options : ['warning_level=2', 'c_std=c99'],
 )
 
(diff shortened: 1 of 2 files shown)


hughsie:
It's a shame we have to ifdef older versions, but I guess LGTM.

jtojnar:
I do not object to this but, if we want to do this, I would like to have an official policy on what meson versions do we intend to support. That way, it will be clear when we can remove this hack.

Are we fine with explicitly supporting just the latest Ubuntu LTS? [Apparently](https://en.wikipedia.org/wiki/Ubuntu_version_history#Version_timeline), 14.04 LTS is still publicly supported for a month. What about other distros like Fedora, or, :nauseated_face::fearful:, CentOS?

superm1:
That's a great point.  Fedora doesn't (currently) offer an LTS, so I think latest Ubuntu LTS is a good line in the sand.

superm1:
Per discussion in https://github.com/hughsie/fwupd/pull/1090 decided to not merge this.  Ubuntu can pick it up downstream in packages.

PR #20, opened 2019-05-08
Make sure to not reuse same GError when ignoring a parsing failure

kalev:
Untested. I inspected the code to see what could be causing gnome-software issue https://gitlab.gnome.org/GNOME/gnome-software/issues/661 and found this. I think it's highly likely to be the root cause.


hughsie:
Are you sure this is the bug? `g_autoptr(GError) error_local = NULL;` is defined at the start of inside the loop; continue should get a new error each time.

kalev:
Ohh, I misread the code. You are right.

PR #23, opened 2019-05-25
build: Update the required version of Meson

bochecha:
Meson helpfully complains about the fact we required 0.46.0 but used
features only available in 0.50.0:

    WARNING: Project targetting '>=0.47.0' but tried to use feature introduced in '0.50.0': install arg in configure_file

This requires updating the CI to Fedora 30, which seems like a good idea anyway.

### DIFF
--- a/contrib/ci/Dockerfile-debian
+++ b/contrib/ci/Dockerfile-debian
@@ -8,9 +8,16 @@ RUN apt-get install -yq --no-install-recommends \
 	libgirepository1.0-dev \
 	libglib2.0-dev \
 	libstemmer-dev \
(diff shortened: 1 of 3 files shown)


bochecha:
The CI currently fails because Debian Sid only has Meson 0.49.2.

However, Debian Experimental has Meson 0.50.1, which should move to Sid soon. (the Debian packager seems to be the upstream developer)

I guess this can wait until then. :slightly_smiling_face:

bochecha:
@hughsie this should be ready for review now, whenever you have a moment. :slightly_smiling_face:

PR #25, opened 2019-06-13
ci: Verify we don't break the ABI

bochecha:
This adds a script which can check for ABI breaks between two Git
revisions:

    $ ./contrib/ci/check-abi.sh ${OLD} ${NEW}

The CI is set up to run it automatically between the tip of the branch
being tested and the last release tag.

---

A couple of notes:

1.  this assumes tags are used only for releases, and compares the ABI of the branch being tested (in a pull request) to the last release tag; is that a fair assumption? Would you prefer comparing to master?
2.  it runs automatically in the CI, but you can also run it manually;
3.  if there's an ABI change libabigail thinks can break apps, it will fail the CI; you could then decide to merge anyway I guess, if you know it isn't an actual breakage?

On that note, libabigail sometimes reports false positives. As I said on IRC, Dodji loves [bug reports](https://sourceware.org/bugzilla/enter_bug.cgi?product=libabigail) about false positives, especially if they come with the two unstripped binaries. :wink:

In addition, there is a suppression mechanism to tell libabigail to ignore changes you know to be safe but that it can't possibly detect as such. I'm happy to help set this up if that's ever needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,14 @@ before_script:
 
 script:
   - docker run -t -v `pwd`:/build libxmlb-${OS} ./contrib/ci/build_and_test.sh
+
+jobs:
+  include:
+    - stage: abi
+      name: "Check for ABI breaks"
+      env:
+        - OS=fedora
+      before_script:
+        - docker build -t libxmlb-${OS} -f contrib/ci/Dockerfile-${OS} .
+      script:
+        - docker run -t -v `pwd`:/build libxmlb-${OS} ./contrib/ci/check-abi $(git describe --abbrev=0 --tags) $(git rev-parse HEAD)
--- a/contrib/ci/Dockerfile-fedora
+++ b/contrib/ci/Dockerfile-fedora
@@ -3,9 +3,11 @@ FROM fedora:30
 RUN dnf -y update
 RUN dnf -y install \
 	gcovr \
+	git-core \
 	glib2-devel \
 	gobject-introspection-devel \
 	gtk-doc \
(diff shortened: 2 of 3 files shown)


hughsie:
I love everything about this, except the shell bit. Could we do it in Python instead? At least my other projects all have the extra helper scripts in python. I'd be happy to be the one to rewrite this if I'm being cheeky.

hughsie:
I love everything about this, except the shell bit. Could we do it in Python instead? At least my other projects all have the extra helper scripts in python. I'd be happy to be the one to rewrite this if I'm being cheeky.

bochecha:
No problem about Python, I already wrote the ABI checker for the Freedesktop Sdk in Python. :slightly_smiling_face:

I'll update.

bochecha:
Now with the script in Python. (heavily "inspired" from the Freedesktop Sdk one :stuck_out_tongue:)

A few additional notes:

1. the ABI check now runs in its own stage, after the other 2 jobs (testing on Debian and Fedora) passed;
2. the script assumes the library is called `libxmlb.so`, if you rename (e.g to `libxmlb-1.0.so` or something like that), it won't work any more;
3. the script tells libabigail to ignore new symbols, because without this the CI would fail every time you add a new symbol. Do you prefer having an explicit failure, so you can review it and decide to merge anyway it because it's just ABI additions? Or are you ok with ignoring added symbols?

hughsie:
> Or are you ok with ignoring added symbols?

I think this is the best approach. Really impressive, good work!

PR #27, opened 2019-06-13
A couple of CI improvements

bochecha:
This pull request makes the CI a bit nicer:

1. it removes an unnecessary operation from the Docker files;
2. it makes the jobs run in parallel on Debian and Fedora;

The most significant change is the latter, because it means testing libxmlb on other distros won't take longer to run, if everything runs in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,12 @@ dist: trusty
 services:
   - docker
 
+env:
+  - OS=debian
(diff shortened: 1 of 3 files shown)


hughsie:
Looks good, just needs rebasing against the new meson stuff. Thanks!

bochecha:
@hughsie rebased.

PR #37, opened 2019-11-20
Wip/hughsie/win32

hughsie:


meson.build
```diff
@@ -107,6 +110,10 @@ installed_test_bindir = join_paths(libexecdir, 'installed-tests', meson.project_
 installed_test_datadir = join_paths(datadir, 'installed-tests', meson.project_name())
 
 gio = dependency('gio-2.0', version : '>= 2.45.8')
```
superm1: how does this get resolved for a Windows build?  Don't you need a gio dll for it?

superm1: I guess `mingw64-glib2`?

hughsie:
> What's the goal with this?

Well, the evening started out with _lets make fwupd compile for BSD_ and ended up dangerously close to having a working `fwupdtool.exe`

> Actually being able to release usable binaries with tagged releases?

For libxmlb certainly not. For fwupd, who knows, but I don't think it's something I want to support. On the fwupd side I found a lot of cases where things worked by accident, but without actually including the correct headers. e.g. the dell plugin calls `sleep(1)` rather than using the more portable `g_usleep()`.

superm1:
> Well, the evening started out with lets make fwupd compile for BSD and ended up dangerously close to having a working fwupdtool.exe

Sounds like a fun evening 🎲 

> For libxmlb certainly not. For fwupd, who knows, but I don't think it's something I want to support. 

Could be a pretty crazy world if it led to people using fwupd cross-platform.

> On the fwupd side I found a lot of cases where things worked by accident, but without actually including the correct headers. e.g. the dell plugin calls sleep(1) rather than using the more portable g_usleep().

Well certainly a lot of the plugins wouldn't make sense outside of Linux though with how dependent upon Linux ioctl or sysfs APIs today.  In some cases there might be replacement Windows APIs and room for `#ifdef WIN32` do this instead, but I would start with a really minimal set of plugins (like only USB).

superm1:
> meson.build:1:0: ERROR: Unknown compiler(s): ['cc', 'gcc', 'clang', 'pgcc', 'icc']

I think that your current CI failure is referring to build compiler, not host compiler.  Meson needs both installed.

It should look something like this:
```
The Meson build system
Version: 0.51.2
Source dir: /home/supermario/src/libxmlb
Build dir: /home/supermario/src/libxmlb/build
Build type: cross build
Project name: libxmlb
Project version: 0.1.14
C compiler for the build machine: ccache cc (gcc 9.2.1 "cc (Ubuntu 9.2.1-16ubuntu1) 9.2.1 20191030")
C compiler for the host machine: /usr/bin/i686-w64-mingw32-gcc (gcc 9.2.0 "i686-w64-mingw32-gcc (GCC) 9.2-win32 20191008")
Host machine cpu family: x86_64
Host machine cpu: i686
Target machine cpu family: x86_64
Target machine cpu: i686
Build machine cpu family: x86_64
Build machine cpu: x86_64
```

hughsie:
> I think that your current CI failure is referring to build compiler, not host compiler.

Doh! Thanks. I've done a few more fixups and most of the tests now pass. The inotify stuff had to be disabled as even the polling fallback didn't seem to work in wine. I suspect it might work better on an actual Windows machine.

src/xb-builder-source-ctx.c
```diff
@@ -112,9 +115,25 @@ xb_builder_source_ctx_get_content_type (XbBuilderSourceCtx *self,
...
-			return g_content_type_guess (priv->filename, buf, bufsz, NULL);
+			content_type = g_content_type_guess (priv->filename, buf, bufsz, NULL);
 	}
-	return g_content_type_guess (priv->filename, NULL, 0, NULL);
+
+	/* either unseekable, or empty */
+	if (content_type == NULL)
+		content_type = g_content_type_guess (priv->filename, NULL, 0, NULL);
+
+#ifdef __MINGW64__
```
superm1: Should this maybe be a `#ifdef WIN32` instead?  So if someone decides to compile this with visual studio instead of mingw64 some day it still works?

hughsie: Is WIN32 defined for a 64 bit build?

superm1: Looks like no; should be `_WIN64`
https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

hughsie: > Is WIN32 defined for a 64 bit build?

The answer to that one is yes, and it's defined in MinGW -- so I've switched to using that.

PR #38, opened 2019-11-29
Wip/hughsie/tmp

hughsie:


meson.build
```diff
@@ -102,12 +102,17 @@ add_project_link_arguments(
...
 )
 
-prefix = get_option('prefix')
-
-libexecdir = join_paths(prefix, get_option('libexecdir'))
-datadir = join_paths(prefix, get_option('datadir'))
-installed_test_bindir = join_paths(libexecdir, 'installed-tests', meson.project_name())
-installed_test_datadir = join_paths(datadir, 'installed-tests', meson.project_name())
+if host_machine.system() == 'windows'
+  libexecdir = '.'
```
superm1: Shouldn't this actually still be get_option? Although in practice I guess we would be using '.'

hughsie: Yes, I guess using get_option does allow us to control this better. I've fixed this, thanks.

PR #40, opened 2019-11-29
Ignore adaptors added with xb_builder_source_add_converter()

hughsie:
There are lots of corner cases I don't want to support with this legacy,
long deprecated API.


hughsie:
> What do you think about putting warnings in the deprecated functions so that any use is noted?

I guess that makes sense, thanks. Fixed.

PR #42, opened 2020-02-24
Wip/hughsie/xb query flag reverse

hughsie:


src/xb-query.c
```diff
@@ -145,6 +146,41 @@ xb_query_set_limit (XbQuery *self, guint limit)
...
+ *
+ * Returns: #XbQueryFlags, default %XB_QUERY_FLAG_NONE
+ *
+ * Since: 0.1.15
+ **/
+XbQueryFlags
+xb_query_get_flags (XbQuery *self)
+{
+	XbQueryPrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (XB_IS_QUERY (self), 0);
```
superm1: Shouldn't this be explicitly `XB_QUERY_FLAG_NONE`?

PR #43, opened 2020-02-27
Add xb_builder_source_add_simple_adapter

djcampello:
This new function allows to specify an output_content_type for an
adapter in order to avoid sniffing mime types when that is not possible.

### DIFF
--- a/src/xb-builder-source.c
+++ b/src/xb-builder-source.c
@@ -36,6 +36,7 @@ typedef struct {
 	XbBuilderSourceAdapterFunc	 func_adapter;
 	gpointer			 user_data;
 	GDestroyNotify			 user_data_free;
+	gboolean			 is_simple;
 } XbBuilderSourceAdapter;
 
 static XbBuilderSourceAdapter *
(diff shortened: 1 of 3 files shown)


src/xb-builder-source.c
```diff
@@ -316,24 +317,26 @@ xb_builder_source_add_converter (XbBuilderSource *self,
...
-			       const gchar *content_types,
-			       XbBuilderSourceAdapterFunc func,
-			       gpointer user_data,
-			       GDestroyNotify user_data_free)
+xb_builder_source_add_simple_adapter (XbBuilderSource *self,
+				      const gchar *content_types,
+				      XbBuilderSourceAdapterFunc func,
+				      gpointer user_data,
+				      GDestroyNotify user_data_free,
+				      const gchar *output_content_type)
```
hughsie: I can't see a use case for the `output_content_type` -- if it's simple it's always going to be `application/xml`, no? You could just set a `gboolean` in `XbBuilderSourceAdapter`

djcampello: Done... Thanks!

PR #45, opened 2020-03-21
tests: Make the build reproducible

bigon:


bigon:
The LGTM test fails due to:

> [2020-03-21 12:41:38] [build] meson.build:1:0: ERROR: Meson version is 0.51.2 but project requires >=0.52.0

hughsie:
I tried and failed to build meson from pip here: https://github.com/hughsie/libxmlb/pull/46 -- ideas very welcome :(

hughsie:
We only have meson 0.49.2 in RHEL 8 -- is there any way of changing this patch to use that version instead?

superm1:
It looks like environment() was added in 0.35.0:
https://github.com/mesonbuild/meson/commit/2a030e33f3c000134fd0d19204133befcae338c2
Maybe we should do a try specifically with RHEL 8's 0.49.2 to see that there is no regression from behavior of `environment()`

hughsie:
I've opened an alternative version here which uses the same trick as libjcat: https://github.com/hughsie/libxmlb/pull/68

PR #46, opened 2020-03-22
trivial: Install meson using pip for LGTM

hughsie:


superm1:
This looks like it actually works properly, should it merge now so that #45 can merge too?

hughsie:
I'm concerned about accidentally depending on newer features that the 0.49.2 we have in RHEL 8.

PR #48, opened 2020-04-02
Don’t allocate opcodes individually

pwithnall:
On a simple test of loading gnome-software and switching between a
couple of categories, this reduces calls to `g_slice_new0()` from around
21 million to around 5 million.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ the text XML data where a lot of the tag content is duplicated, but can actually
 be larger than the original XML file. This isn't important; the fast query speed
 and the ability to mmap strings without copies more than makes up for the larger
 on-disk size. If you want to compress your XML, this library probably isn't for
-you -- just use gzip -- its gives you an almost a perfect compression ratio for
+you -- just use gzip -- it gives you an almost a perfect compression ratio for
 data like this.
 
 For example:
--- a/meson.build
+++ b/meson.build
@@ -27,7 +27,7 @@ configinc = include_directories('.')
 
 # get suported warning flags
 warning_flags = [
-  '-Waggregate-return',
+  '-Wno-aggregate-return',
   '-Wunused',
   '-Warray-bounds',
   '-Wcast-align',
--- a/src/libxmlb.map
+++ b/src/libxmlb.map
@@ -95,12 +95,6 @@ LIBXMLB_0.1.1 {
     xb_machine_parse;
     xb_machine_run;
     xb_machine_set_debug_flags;
-    xb_machine_stack_pop;
-    xb_machine_stack_push;
-    xb_machine_stack_push_integer;
-    xb_machine_stack_push_text;
-    xb_machine_stack_push_text_static;
-    xb_machine_stack_push_text_steal;
     xb_opcode_cmp_str;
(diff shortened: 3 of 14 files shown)


pwithnall:
I‘m putting this here for discussion — it’s not ready to merge. It is in a state where gnome-software will run, with better performance, though.

Essentially, this changes `XbOpcode` to be allocated directly inside the `XbStack` rather than as individual heap allocations. This is good, because opcodes are not really passed around outside stacks, and are created and destroyed very frequently. (In my simple test of loading gnome-software and changing between categories a few times, `xb_stack_push()` is called 30 million times.)

It reduces the number of calls to `g_slice_alloc0()` from around 21 million to around 5 million. Further improvements may be possible with a bit more work.

If the general approach looks reasonable, I can tidy it up and get it ready for actual review and merging. I assume that I can’t break libxmlb’s API/ABI (currently, the patch does, but only for expediency; I can make sure it doesn’t by the time it’s ready for review). If it *is* actually OK to break API, then that would make things a lot easier, and I can drop a load of compatibility code from `XbOpcode` for supporting allocated opcodes.

hughsie:
Sounds like a nice thing to try. I don't think we need to be too precious with the internal API, even if exported. If we have to make changes to fwupd or gnome-software then that would NAK it, but stubbing out internal functions as NOP for the ABI checker is perfectly fine. I guess I care more about raw speed than the number of allocs, although I'm guessing there's some speedup not calling into the allocator so much. I'd certainly prefer a cleaner internal API than being precious about compat.

hughsie:
I've added you as a collaborator. If you can move this branch to origin/wip/pwithnall/opcode-allocations I can jump in and fix up any style issues for you.

pwithnall:
Thanks! Currently in the middle of a release crunch at the moment, but I will get back to this to push it forward in the next few days.

pwithnall:
> I've added you as a collaborator. If you can move this branch to origin/wip/pwithnall/opcode-allocations I can jump in and fix up any style issues for you.

Moving the branch would mean closing this PR and creating a new one, which isn’t ideal. I’ve ticked the “Allow edits from maintainers” tickbox, so you [should be able to push directly to the branch on my fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

pwithnall:
I’m starting to rework this now. So much for “get back to this … in the next few days”, sorry :(

pwithnall:
I’ve pushed my progress so far, so it’s saved somewhere. Next step is getting the tests working again (I had disabled them before), and working out the API tests. This isn’t ready for review yet. I’ll continue tomorrow.

pwithnall:
Right, it’s ready for initial review! I couldn’t really find a way to split out the bulk of the changes, so it’s a single mega-commit. There are probably some rough edges.

I’ve arbitrarily assumed that new API and deprecated API are being given version 0.2.0. That can change.

Grab me on IRC next week if you want to talk through the changes in real time, if that helps!

pwithnall:
Rebased and merged as #49.

PR #49, opened 2020-04-27
Do not allocate opcodes individually

hughsie:
On a simple test of loading gnome-software and switching between a couple of
categories, this reduces calls to `g_slice_new0()` from around 21 million to
around 5 million.

It reduces the time to show (refine) the ‘Games’ category from around 9s to
around 5s the first time it’s shown (in a very unrigorous test).

This makes `XbOpcode` stack-only, and removes all API which requires it to be
heap-allocated. For the most part, this only requires mechanical changes in the
rest of the codebase to operate on stack-allocated opcodes, removing one layer
of indirection in some places.

The largest other change needed was to rewrite the optimiser to avoid using
`GArray`, as tracking the ownership of `XbOpcode`s through several different
`GArray`s was getting too complex. The new implementation of the optimiser
works as before, except it only uses one helper stack rather than a helper
stack and two temporary arrays. This brings its own inherent performance
benefits, as around 5 million XPath queries are done when gnome-software starts
up, so making the optimiser run quickly is important.

This ABI breaks API and ABI, and thus also bumps soname.

### DIFF
--- a/contrib/libxmlb.spec.in
+++ b/contrib/libxmlb.spec.in
@@ -24,10 +24,6 @@ BuildRequires: python-setuptools
 # needed for the self tests
 BuildRequires: shared-mime-info
 
-%if 0%{?rhel} < 8
-BuildRequires: python34
-%endif
-
 Requires: glib2%{?_isa} >= %{glib2_version}
 Requires: shared-mime-info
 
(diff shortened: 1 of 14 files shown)


pwithnall:
Those two new commits (245d6e1 and 080961d) LGTM.

hughsie:
@pwithnall I'm basically happy with this now. Can you see any other API we should remove based on your refactoring?

hughsie:
(I forgot to say, with this branch installed the fwupd and gnome-software tests still pass too)

pwithnall:
> @pwithnall I'm basically happy with this now. Can you see any other API we should remove based on your refactoring?

Nope, LGTM!

pwithnall:
Thanks a lot for taking the time to review this, tweak it, and get it merged so quickly! :)

PR #52, opened 2020-05-29
xb-builder-source-ctx: Store a GFile for the source too

pwithnall:
As well as an input stream for reading the source, store a `GFile`
pointing to the on-disk file which backs the source, if possible.

This allows callers to load the source data directly from disk, which
may be more efficient than streaming it in, and will almost certainly be
more efficient than allocating a 128MB GBytes buffer and reading it all
at once. For example, a caller could memory map moderately sized files.

This adds new API, and breaks some internal API.

### DIFF
--- a/src/xb-builder-source-ctx-private.h
+++ b/src/xb-builder-source-ctx-private.h
@@ -12,7 +12,8 @@
 
 G_BEGIN_DECLS
 
-XbBuilderSourceCtx *xb_builder_source_ctx_new		(GInputStream		*istream);
+XbBuilderSourceCtx *xb_builder_source_ctx_new		(GFile			*file,
+							 GInputStream		*istream);
 void		 xb_builder_source_ctx_set_filename	(XbBuilderSourceCtx	*self,
 							 const gchar		*filename);
 gchar		*xb_builder_source_ctx_get_content_type	(XbBuilderSourceCtx	*self,
--- a/src/xb-builder-source-ctx.c
+++ b/src/xb-builder-source-ctx.c
@@ -85,7 +85,7 @@ _g_input_stream_read_bytes_in_chunks (GInputStream *stream,
  *
  * Returns the data currently being processed.
  *
- * Returns: (transfer none): a #GInputStream
+ * Returns: (transfer full): a #GBytes
  *
  * Since: 0.1.7
  **/
@@ -110,7 +110,7 @@ xb_builder_source_ctx_get_bytes (XbBuilderSourceCtx *self,
  *
  * Returns the basename of the file currently being processed.
  *
- * Returns: a basename, or %NULL if unset
+ * Returns: (transfer none) (nullable): a basename, or %NULL if unset
  *
  * Since: 0.1.7
  **/
@@ -125,11 +125,13 @@ xb_builder_source_ctx_get_filename (XbBuilderSourceCtx *self)
 /**
  * xb_builder_source_ctx_get_content_type:
  * @self: a #XbBuilderSourceCtx
+ * @cancellable: a #GCancellable, or %NULL
+ * @error: the #GError, or %NULL
  *
  * Returns the content type of the input stream currently being
  * processed.
  *
- * Returns: (transfer full): a content type (e.g. `application/x-desktop`), or %NULL
+ * Returns: (transfer full): a content type (e.g. `application/x-desktop`), or %NULL on error
  *
  * Since: 0.1.7
  **/
@@ -192,7 +194,7 @@ xb_builder_source_ctx_class_init (XbBuilderSourceCtxClass *klass)
 
 /**
  * xb_builder_source_ctx_new:
- * @element: An element name, e.g. "component"
+ * @istream: (transfer none): Input stream to load the source from
  *
  * Creates a new builder source_ctx.
  *
--- a/src/xb-builder-source-ctx.c
+++ b/src/xb-builder-source-ctx.c
@@ -86,6 +86,10 @@ _g_input_stream_read_bytes_in_chunks (GInputStream *stream,
  *
  * Returns the data currently being processed.
  *
+ * If the #XbBuilderSourceCtx is backed by a file, the returned #GBytes may be
+ * memory-mapped, and the backing file must not be modified until the #GBytes is
+ * destroyed.
+ *
  * Returns: (transfer full): a #GBytes
  *
  * Since: 0.1.7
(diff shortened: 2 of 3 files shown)


pwithnall:
Used in gnome-software here: https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/464

pwithnall:
Rebased and updated for the 0.3.0 release. No other changes, since the discussion in https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/464 didn’t seem to go anywhere that I could figure out.

It would be good to get this (or an alternative) fix for the problem in https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/464 into the libxmlb 0.3.0 release.

hughsie:
Alternate proposal in https://github.com/hughsie/libxmlb/pull/76

pwithnall:
I think memory mapping is the right way to go for this, as that avoids all complication about heap allocation or growing a read buffer to the right size.

So from my POV this MR is still valid. If you’re working on this stuff now, I’m happy to iterate on the MR (and [the accompanying gnome-software one](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/464) right now to get this resolved.

hughsie:
The issue here is that XbBuilderSourceCtx is supposed to basically wrap a GInputStream -- it doesn't need to be backed by a GFile at all -- it could just be a lump of memory or the output of a decompressor. Could we use xb_builder_source_ctx_get_stream() and pass that to as_app_parse_input_stream()? I know the latter doesn't exist but I think that's the proper way to fix this if we don't want to either leak out the GFile or create a buffered blob of the stream.

pwithnall:
> The issue here is that XbBuilderSourceCtx is supposed to basically wrap a GInputStream -- it doesn't need to be backed by a GFile at all -- it could just be a lump of memory or the output of a decompressor.

The PR does take that into account: the `GFile` is nullable, and `get_bytes()` continues to work as before for streams which aren’t mmappable. mmapping is only an optimisation.

I’ve just pushed an update to this PR which implements mmapping within `get_bytes()` where possible. That means the [changes in gnome-software](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/464) are no longer needed (I don’t know why I didn’t do it this way in the first place). 

With this PR applied, setup on the `appstream` plugin in gnome-software takes 1.7s (as reported by the gnome-software logging for a cold start of gnome-software, zero repeats of the test). Without this PR (but with today’s libxmlb master, i.e. with your `_g_input_stream_read_bytes_in_chunks()` changes) it takes 1.8s. Without either (i.e. on libxmlb commit 6649575b61bdb006dae5680af916e37001baec4b), it takes 9.9s. So your chunk-loading changes are a huge win anyway. :partying_face: 

I could make further changes to this PR to drop the new `xb_builder_source_ctx_get_file()` public API since it’s no longer needed. The (private) changes to `xb_builder_source_ctx_new()` need to stay, though.

src/libxmlb.map
```diff
@@ -213,6 +213,7 @@ LIBXMLB_0.2.0 {
 
 LIBXMLB_0.3.0 {
   global:
+    xb_builder_source_ctx_get_file;
```
hughsie, line 216: i think nuke this

src/xb-builder-source-ctx-private.h
```diff
@@ -12,7 +12,8 @@
 
 G_BEGIN_DECLS
 
-XbBuilderSourceCtx *xb_builder_source_ctx_new		(GInputStream		*istream);
+XbBuilderSourceCtx *xb_builder_source_ctx_new		(GFile			*file,
+							 GInputStream		*istream);
```
hughsie, line 16: Maybe we add xb_builder_source_ctx_file_new() instead?

src/xb-builder-source-ctx.c
```diff
@@ -98,6 +123,23 @@ xb_builder_source_ctx_get_bytes (XbBuilderSourceCtx *self,
...
+	 * be modified during the lifetime of the returned #GBytes. */
+	if (priv->file != NULL) {
+		g_autoptr(GMappedFile) mapped_file = NULL;
+		g_autoptr(GError) error_local = NULL;
+		const gchar *filename = g_file_peek_path (priv->file);
+
+		mapped_file = g_mapped_file_new (filename, FALSE, &error_local);
+		if (mapped_file == NULL)
+			g_debug ("Failed to mmap ‘%s’: %s. Falling back to reading file.",
+				 filename, error_local->message);
```
hughsie, line 138: i dont think we need the g_debug personally.. but i like the mmap

src/xb-builder-source-ctx-private.h
pwithnall: Are you sure? I’d agree if `file` and `istream` were mutually exclusive, but `istream` is always required here, and `file` is optional (nullable). Making them separate functions would complicate the call in `xb-builder-source.c` — it’d go from
```
g_autoptr(XbBuilderSourceCtx) ctx = xb_builder_source_ctx_new (file, priv->istream);
```
to
```
g_autoptr(XbBuilderSourceCtx) ctx = NULL;
if (file != NULL)
  ctx = xb_builder_source_ctx_file_new (file, priv->istream);
else
  ctx = xb_builder_source_ctx_new (priv->istream);
```

hughsie: > Are you sure

I'm never sure! Maybe you're right.

src/xb-builder-source.c
```diff
@@ -471,6 +474,7 @@ xb_builder_source_get_istream (XbBuilderSource *self,
 			return NULL;
 		xb_builder_source_remove_last_extension (basename);
 		g_set_object (&priv->istream, istream_tmp);
+		file = NULL;
```
hughsie, line 477: this makes sense, but can you add a comment explaining how it's only valid for the "outer" one please -- then LGTM.

pwithnall: Good suggestion, done, thanks

src/xb-builder-source-ctx.c
```diff
@@ -98,6 +103,20 @@ xb_builder_source_ctx_get_bytes (XbBuilderSourceCtx *self,
...
+
+	/* Try mmap()ing the file first, as that avoids buffer allocation.
+	 * Note that this imposes the restriction that the backing file must not
+	 * be modified during the lifetime of the returned #GBytes. */
+	if (priv->file != NULL) {
+		g_autoptr(GMappedFile) mapped_file = NULL;
+		g_autoptr(GError) error_local = NULL;
+		const gchar *filename = g_file_peek_path (priv->file);
+
+		mapped_file = g_mapped_file_new (filename, FALSE, &error_local);
```
hughsie, line 115: I think nuke error_local not, i.e. -> NULL.

hughsie (approved):
tiny nit then LGTM, thanks!

src/xb-builder-source-ctx.c
pwithnall: Oops, yes!

PR #53, opened 2020-05-29
Eliminate unnecessary GPtrArray allocations in XbBuilderNode

pwithnall:
See the commit messages. tl;dr: This drops 17.4MB of heap allocations of empty `GPtrArray`s which never get used. (That’s about 833000 `GPtrArray` instances.) This reduces the peak memory consumption of gnome-software when it’s rebuilding its xmlb caches.

### DIFF
--- a/src/xb-builder-node.c
+++ b/src/xb-builder-node.c
@@ -28,11 +28,16 @@ typedef struct {
 	guint32			 tail_idx;
 	XbBuilderNode		*parent;	/* noref */
 	GPtrArray		*children;	/* of XbBuilderNode */
-	GPtrArray		*attrs;		/* of XbBuilderNodeAttr */
+
+	/* Around 80% of all XML nodes have zero attributes, so this array is only
+	 * allocated if it’s non-empty. %NULL means an empty array. */
+	GPtrArray		*attrs;		/* (element-type XbBuilderNodeAttr) (nullable) */
 } XbBuilderNodePrivate;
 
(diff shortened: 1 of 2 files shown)


pwithnall:
The ABI check is failing, I’m guessing because it hasn’t been updated since the development version number was bumped to 0.2.0?

src/xb-builder-node.c
```diff
@@ -27,13 +27,21 @@ typedef struct {
 	gchar			*tail;
 	guint32			 tail_idx;
 	XbBuilderNode		*parent;	/* noref */
-	GPtrArray		*children;	/* of XbBuilderNode */
-	GPtrArray		*attrs;		/* of XbBuilderNodeAttr */
+
+	/* Around 87% all XML nodes have zero children, so this array is only
```
wjt, line 31:
```suggestion
	/* Around 87% of all XML nodes have zero children, so this array is only
```

src/xb-builder-node.c
```diff
@@ -27,13 +27,21 @@ typedef struct {
...
+	 * allocated if it’s non-empty. %NULL means an empty array. */
+	GPtrArray		*children;	/* (element-type XbBuilderNode) (nullable) */
+
+	/* Around 80% all XML nodes have zero attributes, so this array is only
```
wjt, line 35:
```suggestion
	/* Around 80% of all XML nodes have zero attributes, so this array is only
```

pwithnall: Oops, fixed. Thanks

PR #54, opened 2020-06-02
xb: Drop redundant parent_instance members from GObject private structs

pwithnall:
The `parent_instance` member is only needed when creating a derived
`GObject` without a separate private data struct (for example, when
using `G_DECLARE_FINAL_TYPE`).

When using `G_DEFINE_TYPE_WITH_PRIVATE` the `MyTypePrivate` struct only
needs to contain the private members, no boilerplate.

This saves around 500KB of memory in the `XbNode`s created by the silo
when opening gnome-software.

### DIFF
--- a/src/xb-builder-fixup.c
+++ b/src/xb-builder-fixup.c
@@ -13,7 +13,6 @@
 #include "xb-builder-fixup-private.h"
 
 typedef struct {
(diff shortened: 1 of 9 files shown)


hughsie:
Doh! Thanks.

PR #56, opened 2020-06-03
xb-builder-node: Avoid using weak pointers

pwithnall:
Weak pointers were previously used to clear the `XbBuilderNode.parent`
pointer from a node to its parent when the parent was destroyed. Each
parent has strong references to its children, however, so has all the
information available to clear their parent pointers manually.

This saves a load of locking within GObject, and 6.8MB of heap memory
used to store the weak pointers and an (otherwise empty) `GDatalist` per
GObject, when generating the libxmlb caches on gnome-software startup.


src/xb-builder-node.c
```diff
@@ -889,6 +905,8 @@ static void
 xb_builder_node_class_init (XbBuilderNodeClass *klass)
 {
 	GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+	object_class->dispose = xb_builder_node_dispose;
```
hughsie, line 909: hah! I didn't think of that. I assume valgrind is still clean in the self tests.

src/xb-builder-node.c
```diff
@@ -872,6 +870,24 @@ xb_builder_node_init (XbBuilderNode *self)
 	priv->children = NULL;  /* only allocated when a child is added */
 }
 
+static void
+xb_builder_node_dispose (GObject *obj)
+{
+	XbBuilderNode *self = XB_BUILDER_NODE (obj);
+	XbBuilderNodePrivate *priv = GET_PRIVATE (self);
+
+	/* Clear all the child nodes’ parent pointers */
```
hughsie, line 879: not title case single line comment please

src/xb-builder-node.c
pwithnall: It is!

PR #57, opened 2020-06-03
Conditionally drop XbSilo node cache

pwithnall:
See the commit messages for reasoning.

tl;dr: This drops 1.8MB of heap memory of `XbNode`s which are used a few times and then never used again, but hang around in the silo’s cache for the lifetime of gnome-software.

### DIFF
--- a/src/xb-silo-query.c
+++ b/src/xb-silo-query.c
@@ -50,11 +50,27 @@ xb_silo_query_node_matches (XbSilo *self,
 	return TRUE;
 }
 
+/**
+ * XbSiloQueryHelperFlags:
+ * @XB_SILO_QUERY_HELPER_NONE: No flags set.
+ * @XB_SILO_QUERY_HELPER_USE_SN: Return #XbSiloNodes as results, rather than
+ *    wrapping them in #XbNode. This assumes that they’ll be wrapped later.
+ *
+ * Flags for #XbSiloQueryHelper.
+ *
+ * Since: 0.2.0
+ */
+typedef enum {
+	XB_SILO_QUERY_HELPER_NONE = 0,
+	XB_SILO_QUERY_HELPER_USE_SN = 1 << 0,
+} XbSiloQueryHelperFlags;
+
 typedef struct {
 	GPtrArray	*sections;	/* of XbQuerySection */
-	GPtrArray	*results;	/* of XbNode */
+	GPtrArray	*results;	/* of XbNode or XbSiloNode (see @flags) */
 	GHashTable	*results_hash;	/* of sn:1 */
 	guint		 limit;
+	XbSiloQueryHelperFlags	 flags;
 	XbSiloQueryData	*query_data;
 } XbSiloQueryHelper;
 
@@ -63,7 +79,10 @@ xb_silo_query_section_add_result (XbSilo *self, XbSiloQueryHelper *helper, XbSil
 {
 	if (g_hash_table_lookup (helper->results_hash, sn) != NULL)
 		return FALSE;
-	g_ptr_array_add (helper->results, xb_silo_node_create (self, sn));
+	if (helper->flags & XB_SILO_QUERY_HELPER_USE_SN)
+		g_ptr_array_add (helper->results, sn);
+	else
+		g_ptr_array_add (helper->results, xb_silo_node_create (self, sn));
 	g_hash_table_add (helper->results_hash, sn);
 	return helper->results->len == helper->limit;
 }
@@ -167,11 +186,13 @@ xb_silo_query_part (XbSilo *self,
 		    GHashTable *results_hash,
 		    XbQuery *query,
 		    XbSiloQueryData *query_data,
+		    XbSiloQueryHelperFlags flags,
 		    GError **error)
 {
 	XbSiloQueryHelper helper = {
 		.results = results,
 		.limit = xb_query_get_limit (query),
+		.flags = flags,
 		.results_hash = results_hash,
 		.query_data = query_data,
 	};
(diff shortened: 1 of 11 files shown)


src/xb-silo.c
```diff
@@ -824,9 +819,6 @@ xb_silo_load_from_file (XbSilo *self,
 
 	/* no longer valid */
 	g_hash_table_remove_all (priv->file_monitors);
-	g_hash_table_remove_all (priv->nodes);
-	g_hash_table_remove_all (priv->strtab_tags);
-	g_clear_pointer (&priv->guid, g_free);
```
hughsie, line 829: this looks wrong

src/xb-silo-query.c
```diff
@@ -181,32 +202,15 @@ xb_silo_query_part (XbSilo *self,
...
- * Returns: (transfer container) (element-type XbNode): results, or %NULL if unfound
- *
- * Since: 0.1.0
- **/
-GPtrArray *
-xb_silo_query_with_root (XbSilo *self, XbNode *n, const gchar *xpath, guint limit, GError **error)
+/* Returns an array with (element-type XbSiloNode) if use_sn is TRUE, and
+ * (element-type XbNode) otherwise. */
+static GPtrArray *
+silo_query_with_root (XbSilo *self, XbNode *n, const gchar *xpath, guint limit, gboolean use_sn, GError **error)
```
hughsie, line 208: could this just take the XbSiloQueryHelperFlags instead?

src/xb-silo-query.c
```diff
@@ -289,6 +299,53 @@ xb_silo_query_with_root (XbSilo *self, XbNode *n, const gchar *xpath, guint limi
...
+ * xb_silo_query_sn_with_root: (skip)
+ * @self: a #XbSilo
+ * @n: (allow-none): a #XbNode
+ * @xpath: an XPath, e.g. `/components/component[@type=desktop]/id[abe.desktop]`
+ * @limit: maximum number of results to return, or 0 for "all"
+ * @error: the #GError, or %NULL
+ *
+ * A version of xb_silo_query_with_root() which returns results as #XbSiloNodes,
+ * rather than as #XbNodes. This is intended to be used internally to save on
+ * intermediate #XbNode allocations.
```
hughsie, line 337: if it's internal, we shouldn't export it

src/xb-silo.c
```diff
@@ -936,21 +928,7 @@ xb_silo_new_from_xml (const gchar *xml, GError **error)
...
-	XbNode *n;
-	XbSiloPrivate *priv = GET_PRIVATE (self);
-	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->nodes_mutex);
-
-	g_return_val_if_fail (locker != NULL, NULL);
-
-	/* does already exist */
-	n = g_hash_table_lookup (priv->nodes, sn);
-	if (n != NULL)
-		return g_object_ref (n);
```
hughsie, line 948: hmm, I added this code for a reason, perhaps for fwupd. Could we conditionalize it with a flag perhaps?

src/xb-silo.c
pwithnall: Oops, I meant to drop that. It’s a change I was playing around with earlier: that code was redundant because `xb_silo_load_from_bytes()` (which is called later in this control flow) does the same thing. But I think it’s clearer to leave them in place.

src/xb-silo-query.c
pwithnall: How do you mean? It’s declared in `xb-silo-query-private.h` and isn’t listed in `libxmlb.map`.

src/xb-silo.c
hughsie: I think I'd prefer a flag, default caching. GNOME Software can set _NO_NODE_CACHE or something.

pwithnall: I’ve read through the code in fwupd in more detail, and I don’t think there’s a neat way to avoid this caching without massively breaking fwupd’s API and reorganising how it passes releases around internally. Basically, the CAB parser is sideloading three bits of binary data onto the `XbNode` for each `release` in its appstream, that’s getting passed around through some other code, and then the engine is reading that sideloaded data and storing it in a `FuRelease` object. If the CAB loader created the `FuRelease` object in the first place, the sideloading wouldn’t be necessary. Essentially the issue is that `XbNode` is being used to expose the XML structure, but also to represent the higher level object (a release) that’s described by the XML. I don’t think it’s possible to be efficient easily with that wide a range of possible use cases. In this case, libxmlb is not “doing one thing and doing it well”. :(

I’ll look into adding a flag. I think it’s a hack, but it seems unavoidable if I’m understanding this fwupd code correctly.

pwithnall: @hughsie, can you explain your reasoning for preferring a flag please? I’d like to make sure I understand the use cases and considerations at work here, in case they affect future optimisation work I might do. :)

hughsie: > can you explain your reasoning for preferring a flag please

That while I think doing things that require lockstep changes to gnome-software and libxmlb is probably an acceptable compromise (e.g. the API break for speed/memory) I think breaking 2/2 consumers is probably going too far.

pwithnall: > breaking 2/2 consumers is probably going too far.

(The original version of this only broke fwupd, not gnome-software.)

I’ve pushed a new version which adds some new flags and properties to the API to make the cache conditional. It’s a bit fiddly how the thread safety interacts with whether the cache is enabled/disabled and I’m not entirely happy with it. Open to suggestions for improving it.

pwithnall: > It’s a bit fiddly how the thread safety interacts with whether the cache is enabled/disabled and I’m not entirely happy with it.

One option for improving that would be to add a construct-only `enable-node-cache` property to `XbBuilder`, which can be used to ensure that the `XbSilo` is always *constructed* with the correct `enable-node-cache` value (it currently can’t be, since it is constructed in `xb_builder_init()`, which is not where the `NO_NODE_CACHE` flag is available). Then `XbSilo:enable-node-cache` can be made construct-only, and the thread safety concerns go away.

pwithnall:
The CI is going to fail until #58 is merged.

src/xb-silo-query.c
hughsie: my mistake, sorry.

pwithnall:
> The CI is going to fail until #58 is merged.

Rebased on master now that #58 has landed.

Issue #50, opened 2020-04-28
XbSilo xb_builder_node_unlink: assertion 'XB_IS_BUILDER_NODE (self)' failed


bigon:
Hello,

On debian unstable (libxmlb  0.1.15) I get the following GLIB critical coming from libxmlb:

> xb_builder_node_unlink: assertion 'XB_IS_BUILDER_NODE (self)' failed

This can be easily reproduce with the following commands:

```shell
$ appstreamcli convert /var/lib/app-info/yaml/ftp.be.debian.org_debian_dists_sid_main_dep11_Components-amd64.yml.gz /tmp/Components-amd64.xml
$ /usr/lib/x86_64-linux-gnu/xb-tool compile /tmp/Components-amd64.xmlb /tmp/Components-amd64.xml
(xb-tool:1900889): XbSilo-CRITICAL **: 19:03:14.826: xb_builder_node_unlink: assertion 'XB_IS_BUILDER_NODE (self)' failed
[...]
```
If somebody want to test and is not running a debian the yaml file can be downloaded at: http://ftp.debian.org/debian/dists/sid/main/dep11/Components-amd64.yml.gz

The [original bug](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/908) has been opened on the gnome-shell project, but the problem seems to be in libxmlb in the end.

hughsie:
@bigon I can't reproduce, but I think it might be something to do with your locale. What are the values of the following environment variables please? `LANGUAGE, LC_ALL, LC_MESSAGES` and `LANG`

ximion:
I can reproduce this issue with
```
LANGUAGE="de"
LANG="de_DE.UTF-8"
``` 
(other values undefined)

hughsie:
> LANGUAGE="de"

Me too! Debugging now!

hughsie:
@ximion -- the XML is invalid:

      <p>
      Zathura can be extended to support multiple document formats using
      plugins. Other features include:
      <ul><li>SyncTeX forward and backward synchronization support.</li><li xml:lang="sv">Stöd för SyncTeX-framåt- och
[...]

PR #58, opened 2020-06-04
xb-builder-node: Don’t crash if priv->children is NULL

pwithnall:
Since commit d12765d0, `priv->children` is loaded lazily, and can be
`NULL`. Commit b7f2509c2b3 (#50) introduced a crash in that case.


hughsie:
Mea cupla. I think we need to get CI working as a prio soon.

PR #61, opened 2020-06-22
Do not use the node cache when building indexes

hughsie:


hughsie:
It's 82kb saved in fwupd...

pwithnall:
LGTM. Don’t know if you want me to merge things like this, or just want the review

hughsie:
> Don’t know if you want me to merge things like this, or just want the review

Either! :) Thanks.

PR #62, opened 2020-06-22
Wip/hughsie/query cache flags

hughsie:


src/xb-silo.c
```diff
@@ -1004,15 +1004,15 @@ xb_silo_new_from_xml (const gchar *xml, GError **error)
...
 XbNode *
-xb_silo_node_create (XbSilo *self, XbSiloNode *sn)
+xb_silo_node_create (XbSilo *self, XbSiloNode *sn, gboolean force_node_cache)
 {
 	XbNode *n;
 	XbSiloPrivate *priv = GET_PRIVATE (self);
 	g_autoptr(GMutexLocker) locker = NULL;
 
 	/* the cache should only be enabled/disabled before threads are
 	 * spawned, so this can be accessed unlocked */
```
pwithnall, line 1014: This comment should probably be amended to specifically refer to `priv->enable_node_cache`, since it’s now ambiguous whether the comment refers to `priv->enable_node_cache` or `force_node_cache`

hughsie: Something like:

    /* the cache should only be enabled/disabled using xb_silo_set_enable_node_cache()
     * before threads are spawned, so this can be accessed unlocked */

?

pwithnall: No, I was thinking more like
```
/* the cache should only be enabled/disabled before threads are
 * spawned, so `priv->enable_node_cache` can be accessed unlocked */
```

hughsie: Ohh I see -- fixed up with your suggestions. Ta.

PR #63, opened 2020-06-22
Export xb_node_get_silo()

hughsie:
We need the silo for a standalone XbNode when building prepared queries using
xb_query_new() and xb_query_new_full().

Split it out to a new header to avoid an #include loop.


pwithnall:
LGTM.

PR #64, opened 2020-07-27
Wip/hughsie/new stack g error

hughsie:


src/xb-machine.c
```diff
@@ -1406,7 +1406,13 @@ xb_machine_func_eq_cb (XbMachine *self,
 	}
 
 	/* should have been checked above */
-	g_assert_not_reached ();
+	g_set_error (error,
```
pwithnall, line 1409: This shouldn’t be necessary. All the appropriate checking should have been done by `xb_machine_check_two_args()` at the start of the function. Something must be wrong there.

src/xb-machine.c
```diff
@@ -1399,53 +1365,44 @@ xb_machine_func_eq_cb (XbMachine *self,
 					xb_opcode_cmp_val_or_str,
 					error))
 		return FALSE;
-
-	xb_machine_stack_pop (self, stack, &op1);
-	xb_machine_stack_pop (self, stack, &op2);
+	if (!xb_machine_stack_pop (self, stack, &op1, error))
+		return FALSE;
+	if (!xb_machine_stack_pop (self, stack, &op2, error))
+		return FALSE;
```
pwithnall, line 1371: I think various other bits of the code attempt to leave the stack unchanged if they error out, so it would probably make sense to add a helper function for popping two opcodes off the stack at once, doing all the stack size checks only once, and making sure that the stack is unchanged if an error is returned.

src/xb-machine.c
hughsie: I never got to this point, I just changed it to match the others. I agree it probably can't be reached, but I didn't like the other `g_assert_not_reached()` just exploding when running under afl.

src/xb-machine.c
hughsie: Yes, probably a nice optimisation for the common case too. Will fix.

src/xb-machine.c
pwithnall: Which one was the `g_assert_not_reached()` which was exploding under afl? That shouldn’t be happening — is an `xb_machine_check_two_args()` (or similar) missing in that case?

hughsie: I believe it was a `ne` with `TEXT:INTE` opcodes.

src/xb-machine.c
```diff
@@ -1490,6 +1490,19 @@ xb_machine_func_ne_cb (XbMachine *self,
 					   error);
 	}
 
+	/* TEXT:INTE */
+	if (xb_opcode_cmp_val (&op2) && xb_opcode_cmp_str (&op1)) {
```
pwithnall, line 1494: Nitpick: it’s a bit weird that the comment is in op1:op2 order, but the tests in the `if` statement are in `op2 && op1` order.

Same below and above.

hughsie: Not a nitpick, that's perfectly valid. Will fix up.

PR #65, opened 2020-08-18
Do not assume g_content_type_guess() always returns valid results

hughsie:
On Chrome OS the database of magic is either incomplete or missing and so
provide the application/gzip fallback manually by sniffing the first two bytes.


superm1:
Hmm as reported at https://chromium-review.googlesource.com/c/chromiumos/third_party/fwupd/+/2354520 this doesn't seem to fully solve the problem.  CC @djcampello 

Maybe `xb_builder_source_ctx_guess_content` needs to handle the magic in the header for XML too.

djcampello:
Sadly, on my experiment it did not seem to solve the gzip issue either (given the error/warning displayed) I'm not sure why octet-stream is still mentioned if this patch is overriding it to gzip

djcampello:
Stepping with gdb... The function gets called twice:

Thread 1 "fwupdtool" hit Breakpoint 1, xb_builder_source_ctx_guess_content (filename=0x555555767990 "chromeos.xml.gz", buf=0x7fffffffcbc0 "\037\213\b\b\364P<_", bufsz=2487)
    at ../libxmlb-0.1.15/src/xb-builder-source-ctx.c:93
(gdb) finish
Run till exit from #0  xb_builder_source_ctx_guess_content (filename=0x555555767990 "chromeos.xml.gz", buf=0x7fffffffcbc0 "\037\213\b\b\364P<_", bufsz=2487)
    at ../libxmlb-0.1.15/src/xb-builder-source-ctx.c:93
0x00007ffff7e14fa0 in xb_builder_source_ctx_get_content_type (self=self@entry=0x5555556ac920, cancellable=<optimized out>, cancellable@entry=0x0, error=<optimized out>, error@entry=0x7fffffffdcf8)
    at ../libxmlb-0.1.15/src/xb-builder-source-ctx.c:145
Value returned is $15 = (gchar *) 0x5555557ab4d0 "application/gzip"

Thread 1 "fwupdtool" hit Breakpoint 1, xb_builder_source_ctx_guess_content (filename=0x5555557ab4d0 "chromeos.xml", buf=0x0, bufsz=0) at ../libxmlb-0.1.15/src/xb-builder-source-ctx.c:93
(gdb) finish
Run till exit from #0  xb_builder_source_ctx_guess_content (filename=0x5555557ab4d0 "chromeos.xml", buf=0x0, bufsz=0) at ../libxmlb-0.1.15/src/xb-builder-source-ctx.c:93
0x00007ffff7e14fa0 in xb_builder_source_ctx_get_content_type (self=self@entry=0x555555696350, cancellable=<optimized out>, cancellable@entry=0x0, error=<optimized out>, error@entry=0x7fffffffdcf8)
    at ../libxmlb-0.1.15/src/xb-builder-source-ctx.c:145
Value returned is $16 = (gchar *) 0x5555556b23c0 "application/octet-stream"

So, it seems adding the xml magic should work

djcampello:
Actually, the second time is called without the buffer contents...

djcampello:
Not sure if this is robust enough... But under the same check for octet-stream, after the gzip magic check, we could have a simple file extension check for xml on filename

hughsie:
> we could have a simple file extension check for xml on filename

I was able to reproduce this myself with the rather drastic step of `sudo mv /usr/share/mime /usr/share/mime.old` -- I've updated the patch so that it works with no mime data at all -- can you try this version again please.

djcampello:
Works! Thanks! I took it in crrev.com/c/2363980

src/xb-common.c
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Returns: a string indicating a guessed content type
+ **/
+gchar *
+xb_content_type_guess (const gchar *filename, const guchar *buf, gsize bufsz)
+{
+	g_autofree gchar *content_type = NULL;
+
+	/* check for bad results, e.g. from Chrome OS */
+	content_type = g_content_type_guess (filename, buf, bufsz, NULL);
```
superm1, line 66: do you know what Windows returns back here?  My thought is there might be an opportunity to remove the `#ifdef` below and just use `xb_content_type_guess_from_fn` in all cases.

PR #66, opened 2020-08-18
Update the header location to reflect the new API

hughsie:


superm1:
The other thing to check - is the GObject introspection still 1.0 ABI?  Or does that bump too?

hughsie:
> Or does that bump too?

Good question, will do some tests.

hughsie:
> will do some tests

It needs a bump too; I've added that as a follow up commit. Is this for Debian where both versions are parallel installable? If so, are there any other files that are conflicting?

superm1:
>It needs a bump too; I've added that as a follow up commit. Is this for Debian where both versions are parallel installable? If so, are there any other files that are conflicting?

That gets the last that I see, thanks!

PR #67, opened 2020-08-19
Rework query binding API

pwithnall:
An initial version of my work to rework how values are bound to queries, so that queries can be cached in a thread-safe way.

This makes all the relevant API breaks and additions, compiles and passes tests, but needs polish (documentation, maybe some convenience APIs, code style, commits need splitting up nicely).

It’s ready for initial review, with the hope that it can be landed for 0.2.0 so libxmlb can freeze API again.

The fundamental change is that values are bound to a stack-allocated `XbValueBindings` object once per query call site, now, rather than being bound to the `XbQuery` object itself. The placeholder `XbOpcode`s are replaced with their bound values in a pre-pass in `xb_machine_run()`.

I’ve deprecated some functions, but could just as easily remove them entirely if that’s the route you want to take. 0.2.0 breaks API already, so that would be fairly uncontroversial.


hughsie:
Dude, I've just done a release with the new API. We can't break API yet again. We'll have to maintain compat stubs.

hughsie:
> so that queries can be cached in a thread-safe way

Do we want to share queries between threads? Or is just that it allows us to cache them in one place?

pwithnall:
> Dude, I've just done a release with the new API. We can't break API yet again. We'll have to maintain compat stubs.

aargh, I missed it by a day! There’s a possibility to label this all as 0.3.0 and release it quickly, suggest that distros not package 0.2.0, and then they still only need to deal with one API break transition?

hughsie:
> suggest that distros not package 0.2.0, and then they still only need to deal with one API break transition

I've already spent 3 hours yesterday building it all for F33 and F34 unfortunately. It's not something I'm super keen to repeat I'm afraid.

> gnome-software already runs the same query from multiple threads

What's the speedup possible out of interest?

pwithnall:
> > suggest that distros not package 0.2.0, and then they still only need to deal with one API break transition
> 
> I've already spent 3 hours yesterday building it all for F33 and F34 unfortunately. It's not something I'm super keen to repeat I'm afraid.

:cry: 

I’ll have to add `xb_node_query_full_full()` functions, or something like that, then.

> > gnome-software already runs the same query from multiple threads
> 
> What's the speedup possible out of interest?

It’ll reduce the number of queries being parsed and turned into `GObject`s on gnome-software startup from about 14000 to less than 100, from the initial profiling I did. I don’t know what kind of a speedup that will be, since there was no way to test this other than by actually writing this PR. There’s scope for it to be good, though: gnome-software spends over 1s of CPU time on appstream queries on startup.

hughsie:
> xb_node_query_full_full

`xb_node_query_with_binding()` perhaps? Quite a lot of the callers of the `xb_node_query` API would have a `NULL` binding anyway.

> There’s scope for it to be good

I'd love to see some numbers :)

pwithnall:
OK, I’m going to leave this for a few days since there’s no longer a rush to make the 0.2.0 release. I’ll come back to it in a bit and rework it to avoid API breaks, although the API additions mean it’ll be 0.3.0 territory rather than 0.2.1 territory.

It’s still ready for initial review for the approach and new API.

hughsie:
> although the API additions mean it’ll be 0.3.0 territory

I'd be happy to add new API in 0.2.1 FWIW...

src/xb-value-bindings.c
```diff
@@ -0,0 +1,233 @@
...
+		gchar *text;
+		guint32 integer;
+	};
+	GDestroyNotify destroy_func;
+} BoundValue;
+
+typedef struct {
+	/* Currently limited to two values since that’s all that any client
+	 * uses. This could be expanded to dynamically allow more in future. */
+	BoundValue values[2];
```
hughsie, line 34: That seems a bit odd to hardcode here

src/xb-value-bindings.h
```diff
@@ -0,0 +1,53 @@
...
+	gpointer dummy1[2];
+	gint dummy2;
+	gpointer dummy3[2];
+	gpointer dummy4[3];
+} XbValueBindings;
+
+/* TODO docs */
+#define XB_VALUE_BINDINGS_INIT() { 0, { NULL, NULL }, 0, { NULL, NULL }, { NULL, } }
+
+void		 xb_value_bindings_init		(XbValueBindings	 *self);
```
hughsie, line 29: I'm not sure I like the _name_. It's a mapping of bound indexes to values, right? More generally, it's thread global context for the query, so maybe something with `Context` in the name?

pwithnall:
Updated to not break API. No other particularly significant changes.Ready for review again.

src/xb-value-bindings.h
pwithnall: I haven’t been able to come up with a better name. It is a mapping of bound indices to values, but I don’t think it’s as general as a context for the query. It doesn’t contain any other contextual data. Subsets of it are copied to new `XbValueBindings` instances for different predicates in the query (see the changes to `xb_silo_query_node_matches()` in “xb-query: Use XbValueBindings in query functions”), which wouldn’t work as well semantically if it was a ‘query context’.

src/xb-value-bindings.c
pwithnall: It’s cutting a corner, but saved me having to implement dynamic sizing for the `XbValueBindings` struct, which isn’t necessary at the moment. There are actually very few queries in gnome-software or fwupd which use bound values, and they only ever use 2 at the most.

src/xb-silo.h
```diff
@@ -1,4 +1,6 @@
-/*
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ * vi:set noexpandtab tabstop=8 shiftwidth=8:
```
hughsie, line 2: Could we just add a .editorconfig instead? https://editorconfig.org/

src/xb-silo.c
```diff
@@ -1530,3 +1538,57 @@ xb_silo_new (void)
...
+				return NULL;
+			result = g_object_ref (query);
+
+			g_hash_table_insert (priv->query_cache, g_strdup (xpath), g_steal_pointer (&query));
+			g_debug ("Caching query ‘%s’ (%p) in silo %p; query cache now has %u entries",
+				 xpath, query, self, g_hash_table_size (priv->query_cache));
+		}
+	}
+
+	return result;
```
hughsie, line 1593: would it be better to ref the result here?

hughsie (changes requested):

src/xb-silo.c
pwithnall: On the `xb_query_new()` path, the query has to be reffed with the write lock held. Moving the ref to the bottom of the function would mean moving the `GRWLockWriterLocker` to the top scope in the function, which I think would make it less clear where the locked sections need to be.

src/xb-silo.h
pwithnall: Done

src/xb-node-query.c
```diff
@@ -58,14 +59,45 @@ xb_node_query (XbNode *self, const gchar *xpath, guint limit, GError **error)
  * Returns: (transfer container) (element-type XbNode): results, or %NULL if unfound
  *
  * Since: 0.1.4
+ * Deprecated: 0.3.0: Use xb_node_query_with_bindings() instead.
```
hughsie, line 62: I don't think we want to deprecate this; I think it's easier to use than xb_node_query_with_bindings(). Basically, in the case of nothing being bound it's harmess and easier to understand.

src/xb-node-query.c
```diff
@@ -84,9 +116,36 @@ xb_node_query_full (XbNode *self, XbQuery *query, GError **error)
  * Returns: (transfer full): a #XbNode, or %NULL if unfound
  *
  * Since: 0.1.11
+ * Deprecated: 0.3.0: Use xb_node_query_first_with_bindings() instead.
```
hughsie, line 119: same?

src/xb-silo-query.c
```diff
@@ -41,11 +45,33 @@ xb_silo_query_node_matches (XbSilo *self,
 	if (section->predicates != NULL) {
 		for (guint i = 0; i < section->predicates->len; i++) {
 			XbStack *opcodes = g_ptr_array_index (section->predicates, i);
-			if (!xb_machine_run (machine, opcodes, result, query_data, error))
+			g_auto(XbValueBindings) predicate_bindings = XB_VALUE_BINDINGS_INIT ();
+			guint predicate_bindings_idx = 0;
+			XbValueBindings *predicate_bindings_ptr = (bindings != NULL) ? &predicate_bindings : NULL;
```
hughsie, line 50: this looks a bit magic... maybe just an `if`?

src/xb-silo-query.c
```diff
@@ -41,11 +45,33 @@ xb_silo_query_node_matches (XbSilo *self,
...
+
+			/* set up the bindings for this predicate */
+			for (guint k = 0; bindings != NULL && k < xb_stack_get_size (opcodes); k++) {
+				XbOpcode *op = xb_stack_peek (opcodes, k);
+				if (xb_opcode_is_binding (op)) {
+					/* ignore errors as they’ll be caught by xb_machine_run() */
+					xb_value_bindings_copy_binding (bindings, bindings_offset + predicate_bindings_idx, &predicate_bindings, predicate_bindings_idx);
```
hughsie, line 57: line too long

src/xb-silo-query.c
```diff
@@ -41,11 +45,33 @@ xb_silo_query_node_matches (XbSilo *self,
...
+					predicate_bindings_idx++;
+				}
+			}
+
+			/* run the predicate; pass NULL for the bindings iff
+			 * (bindings == NULL), as that means we’ve been called
+			 * with pre-0.3.0-style pre-bound values */
+			if (!xb_machine_run_with_bindings (machine, opcodes, predicate_bindings_ptr, result, query_data, error))
```
hughsie, line 65: line too long again

src/xb-silo.h
```diff
@@ -97,4 +97,10 @@ gboolean	 xb_silo_get_enable_node_cache		(XbSilo		*self);
 void		 xb_silo_set_enable_node_cache		(XbSilo		*self,
 							 gboolean	 enable_node_cache);
 
+#include "xb-query.h"
+
+XbQuery		*xb_silo_lookup_query			(XbSilo		*self,
```
hughsie, line 102: do we need to export this?

src/xb-value-bindings.c
hughsie: I guess that's pragmatic. Is it in the docs anywhere? Maybe we could make it say 8 instead as I can think offhand of a place in fwupd that I could use this with three bound values.

src/xb-value-bindings.c
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * Since: 0.3.0
+ */
+void
+xb_value_bindings_init (XbValueBindings *self)
+{
+	RealValueBindings *_self = (RealValueBindings *) self;
+
+	_self->values[0].kind = KIND_NONE;
+	_self->values[1].kind = KIND_NONE;
```
hughsie, line 64: can we add a loop with a MAX_VALUES type thing? then we only have to change a constant at the top in the future.

src/xb-value-bindings.h
hughsie: > I haven’t been able to come up with a better name

Me either. Lets stick to this for now.

src/xb-node-query.c
pwithnall: My thinking about deprecating as much as possible is that there are currently 13 public `*_query*` functions in libxmlb, and (imo) it’s already hard enough to work out which one you should be using. I was hoping to consolidate that a bit or, at least, not make it worse.

> Basically, in the case of nothing being bound it's harmess and easier to understand.

That is, however, true.

src/xb-silo-query.c
pwithnall: As in the following?
```
XbValueBindings *predicate_bindings_ptr = NULL;
if (bindings != NULL)
  predicate_bindings_ptr = &predicate_bindings;
```

Not sure that’s any clearer, but I may be misunderstanding what you mean.

src/xb-silo.h
pwithnall: It’s not needed for any of the gnome-software or fwupd changes I’ve submitted so far, but I have a local gnome-software branch which uses it to cache the query instances which need values binding to them.

It basically exposes the query cache from the `XbSilo` publicly so that applications can get an `XbQuery` out of the silo to bind things to before querying using one of the new `query_with_bindings()` APIs. Those APIs all take an `XbQuery`, so the caller would otherwise either need to create their own cache, or create a new query each time with `xb_query_new()` (not performant).

Requiring that callers create and maintain their own cache is a viable alternative. However, when I tried it in gnome-software, it meant carrying the cache pointer around alongside the `XbSilo` everywhere, which was quite an invasive change.

I guess it depends on whether you think the `XbSilo` should, conceptually, have a cache for the queries it’s executed, or whether that should be separate.

There’s also a separate question about whether the `XbSilo` should cache *every* query it parses, or whether there should be a flag for callers to say “don’t cache the `XbQuery` which results from parsing this XPath string” because, say, the caller knows that query is never going to be used again.

src/xb-value-bindings.c
pwithnall: I was thinking that rather than documenting it, we’d just fix it to expand dynamically if someone hits the limit. Which approach would you prefer?

pwithnall: I’ve expanded it to 4 values (from 2), but don’t want to go as far as 8 because that will make the struct absolutely massive. There’s now plenty of room in there for rearranging things in future to use dynamic values, without breaking ABI.

src/xb-node-query.c
hughsie: Ya, it's a bit chaotic. I guess it's hard because most of the time it's "I want the cdata from this one node" rather than create query, bind values, execute query, get results etc. I think if anything needs to hit the cutting room floor, I'd rather it was the silo query functions, and we can just use the node-query in most cases and XbQuery in anything more complex.

src/xb-silo-query.c
hughsie: ya, i think that's much better. using the `?` always makes my brain hurt and I'm sure it all gets compiled to the same thing anyway.

src/xb-silo.h
hughsie: > it meant carrying the cache pointer around alongside the XbSilo everywhere

Ya, that sounds sucky.

> there should be a flag for callers to say “don’t cache the XbQuery

I very think we need this, else people shout at the memory use in fwupd again.

pwithnall: > > there should be a flag for callers to say “don’t cache the XbQuery
> 
> I very think we need this, else people shout at the memory use in fwupd again.

I just looked at doing this, and I don’t think it will work: in all the APIs where the query is provided as an XPath string, there is no flags argument. In all the other cases, the caller is providing an `XbQuery`, so might as well explicitly call `xb_silo_lookup_cache()` if they want to use the cache.

How does it sound for me to disable caching by default in this MR, and amend the gnome-software/fwupd MRs to use `xb_silo_lookup_cache()` where queries need caching (and continue to use `xb_query_new()` otherwise).

pwithnall: I’ve pushed changes to fix the other review comments, but I won’t address this one until you’ve replied on it, since it’s a more involved change.

hughsie: > in all the APIs where the query is provided as an XPath string, there is no flags argument

Is there any way to encode flags into the query string? It's not awesome, but what about something like `/components/component#cache=true`? Or we set the default for the _silo_?

> How does it sound for me to disable caching by default in this MR, and amend the gnome-software/fwupd MRs to use xb_silo_lookup_cache() where queries need caching (and continue to use xb_query_new() otherwise).

Yes, that works too. At least from a fwupd point of view it's sitting there idle 99.99% of the time and RSS usage is more important than saving a few ms on startup.

pwithnall: OK, I’ve disabled caching by default.

I’ve also pushed a change which drops the `GError**` argument from `xb_silo_lookup_cache()`, on the basis that XPath queries should always be valid. This significantly simplifies the error handling in gnome-software (see the commit message for 2b4f58c23bc53c9940efc5eaa81561ebb7574c8b).

Ready for review again. My recent changes are all pushed as individual fixups, so you can easily see what’s changed since the last review.

pwithnall: Aaargh, I’ve just had a horrible thought that `xb_query_set_flags()` and `xb_query_set_limit()` aren’t thread-safe either. If a cached query is simultaneously run from two different threads, either of those functions could be called with different values from each thread.

I think I might need to rework `XbValueBindings` to be `XbQueryContext` and to contain the flags and limit after all.

Watch this space.

pwithnall: OK, updated. I’ve kept `XbValueBindings` (since they’re needed for the `XbMachine`), and added an `XbQueryContext` which wraps them in all the query APIs, and also contains the limit and flags. I’ve added an additional commit which deprecates `xb_query_{get,set}_{limit,flags}()`.

Ready for review again!

src/xb-node-query.c
```diff
@@ -87,15 +121,51 @@ xb_node_query_full (XbNode *self, XbQuery *query, GError **error)
...
+		old_limit = xb_query_context_get_limit (context);
+		xb_query_context_set_limit (context, 1);
+	}
+
 	/* nodes don't have to include themselves as part of the query */
-	results = xb_silo_query_with_root_full (xb_node_get_silo (self), self, query, error);
+	results = xb_silo_query_with_root_full (xb_node_get_silo (self), self, query, context, error);
+
+	if (context != NULL)
+		xb_query_context_set_limit (context, old_limit);
```
hughsie, line 167: this smells fishy, no?

src/xb-query.c
```diff
@@ -486,7 +507,7 @@ xb_query_new_full (XbSilo *silo, const gchar *xpath, XbQueryFlags flags, GError
 	g_return_val_if_fail (error == NULL || *error == NULL, NULL);
 
 	/* create */
-	priv->silo = g_object_ref (silo);
+	priv->silo = silo;  /* don’t take a reference, otherwise we get refcount loops with cached queries from xb_silo_lookup_query() */
```
hughsie, line 510: should we set a weak pointer thing?

src/xb-silo-query.c
```diff
@@ -371,6 +421,9 @@ _g_ptr_array_reverse (GPtrArray *array)
  * @self: a #XbSilo
  * @n: (allow-none): a #XbNode
  * @query: an #XbQuery
+ * @contextc: (nullable) (transfer none): context including values bound to opcodes of type
```
hughsie, line 424: s/contextc/context

src/xb-silo-query.h
```diff
@@ -18,15 +19,27 @@ GPtrArray	*xb_silo_query			(XbSilo		*self,
...
 						 GError		**error);
+
 XbNode		*xb_silo_query_first_full	(XbSilo		*self,
 						 XbQuery	*query,
 						 GError		**error);
+XbNode		*xb_silo_query_first_with_context(XbSilo	*self,
+						 XbQuery	*query,
+						 XbQueryContext	*context,
+						 GError		**error);
+
```
hughsie, line 42: I'm wondering if we need this at all; if you're using xb_silo_query_with_context() rather than xb_silo_query_full() then it's just one _more_ line to set the limit.

src/xb-silo.h
hughsie: I think this version is much better; good catch on the xb_query_set_flags() thread safety issue. I'm a bit confused why we need the QueryContext (which I like) and also the XbValueBindings object too. Don't they have the same lifetime and basic purpose?

pwithnall: > Don't they have the same lifetime and basic purpose?

Not quite. The `XbQueryContext` is associated with a query, but the `XbValueBindings` is associated with a query and the `XbMachine` which runs for it. The `XbValueBindings` are copied/subsetted in a few places as different parts of the query are run. In my first attempt at this last night, `XbQueryContext` and `XbValueBindings` were the same object, and passing it (including the limit and flags) down to `XbMachine` was an awkward layering violation.

src/xb-silo-query.h
pwithnall: I’m afraid we do, because it’s got a different return type. I think that’s worth the additional API.

src/xb-query.c
pwithnall: The `->silo` member is actually only ever used during parsing. I’ll split it out into a temporary parsing context.

src/xb-silo-query.h
hughsie: okay, fair enough, ta.

src/xb-query.c
hughsie: Yes, I think that's best, thanks.

src/xb-silo.h
hughsie: > down to XbMachine was an awkward layering violation

Fair enough, thanks for the explanation. Can we put some of that text in the code for my future self please? Ta.

src/xb-node-query.c
pwithnall: It does. I was struggling to find a way to make it better. I’ve pushed a commit which might improve things — see what you think.

src/xb-silo.h
pwithnall: Pushed another fixup commit to add a comment about that.

pwithnall:
Ready for review again; I think we crossed paths slightly with my latest updates and your latest review comments. I think I’ve addressed everything now, but let me know if you think I’ve missed something.

src/xb-silo-query.c
```diff
@@ -190,23 +232,29 @@ xb_silo_query_part (XbSilo *self,
 		    GPtrArray *results,
 		    GHashTable *results_hash,
 		    XbQuery *query,
+		    XbQueryContext *context,
+		    gboolean first_result_only,
```
hughsie, line 236: could this be in XbSiloQueryHelperFlags?

src/xb-value-bindings.h
```diff
@@ -0,0 +1,86 @@
...
+typedef struct {
+	gint dummy0;
+	gpointer dummy1[2];
+	gint dummy2;
+	gpointer dummy3[2];
+	gint dummy4;
+	gpointer dummy5[2];
+	gint dummy6;
+	gpointer dummy7[2];
+	gpointer dummy8[3];
```
hughsie, line 35: this looks pretty l33t -- too l33t for me :)

src/xb-silo-query.c
pwithnall: Not easily, that would mean exposing `XbSiloQueryHelperFlags` in `xb-silo-query-private.h` and shifting a lot of stuff around.

At this point I’d really like to just get these changes merged, and then further cleanups like this can be done as-needed! The changes in this commit are all in private APIs, so there’s no problem with API breaks if we do changes there in future.

src/xb-value-bindings.h
pwithnall: There’s no real alternative that I know of. The idea is that the structs in `xb-value-bindings.c` and `.h` mirror each other in terms of size and alignment, but the struct members are obscured (‘dummy’) in the public version, so that people can’t access the struct members directly. Because the struct is public, though, they can still stack-allocate it for speed. The internal C code casts from the public (‘dummy’) struct to the private (real) struct before interacting with the real struct members.

It’s not an uncommon pattern, and is used in various places in GLib, such as `GHashTableIter`.

src/xb-silo-query.c
hughsie: Point taken.

src/xb-value-bindings.h
hughsie: Don't we need a /*< private >*/ sprinkled in tho?

pwithnall: Yes, good suggestion. And for `XbQueryContext`, which uses the same approach. I’ll do that later and then rebase everything ready to (maybe?) be merged. Do you have any other review comments?

hughsie: > Do you have any other review comments?

Nope! LGTM after that, thanks.

pwithnall: OK, done for both, and I’ve squashed all the fixups in.

hughsie: MERGE! :) Thanks.

hughsie (approved):

pwithnall:
:partying_face:  Thanks for the reviews!

PR #68, opened 2020-08-19
Make the build reproducible

hughsie:
Based on a patch by Laurent Bigonville <bigon@bigon.be>, many thanks.


hughsie:
@bigon -- does this work for you?

PR #76, opened 2021-01-15
Do not allocate a 128Mb buffer for each xb_builder_source_ctx_get_bytes()

hughsie:
Instead read the stream in chunks of 32Kb; most sources being smaller than this.

### DIFF
--- a/src/xb-builder-source-ctx.c
+++ b/src/xb-builder-source-ctx.c
@@ -40,6 +40,43 @@ xb_builder_source_ctx_get_stream (XbBuilderSourceCtx *self)
 	return priv->istream;
 }
 
+static GBytes *
+_g_input_stream_read_bytes_in_chunks (GInputStream *stream,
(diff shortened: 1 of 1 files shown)


superm1 (approved):

pwithnall:
Have you profiled this? Some appstream files are 11MB large, which means a number of `g_byte_array_append()` calls to grow the final buffer to that size. Each append call means a `memcpy()`, which are *very* expensive.

hughsie:
> Have you profiled this

Admittedly no, I forgot we used this for AppStream as well as desktop/appdata. Maybe the chunk size could start at say 20Mb and double each reallocation? It's a ton less than 128MB...

pwithnall:
I think the only way to know the right buffer size is to profile, but I don’t think there’s much point in spending time profile-bikeshedding on that if we can fix it in a better way overall with #52.

PR #78, opened 2021-01-21
Wip/hughsie/utf8

hughsie:


hughsie:
This is the bare minimum. I've spent about 4 hours today working on a branch that does the tokenization at silo built time, storing the normalized tokens in the silo. That branch is full of hacks (but works) and I'll continue on that tomorrow.

src/xb-machine.c
```diff
@@ -1796,7 +1796,7 @@ xb_machine_func_lower_cb (XbMachine *self,
 
 	/* TEXT */
 	return xb_machine_stack_push_text_steal (self, stack,
-						 g_ascii_strdown (xb_opcode_get_str (&op), -1),
+						 g_utf8_strdown (xb_opcode_get_str (&op), -1),
```
pwithnall, line 1799: Looking at how `g_utf8_strdown()` is [implemented in GLib](https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/guniprop.c#L1126), its behaviour differs depending on whether the current locale is ‘normal’, Lithuanian or Turkic (for example, Turkish or Uzbek). It also won’t work reliably if the input is not valid UTF-8.

Those are fairly cornery corner cases, so I don’t think these comments need to be addressed in this MR, but you may want to do the following in a follow-up:
 * Ensure that libxmlb rejects XML inputs which are invalid UTF-8 early on (and [test this](https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt))
 * Consider whether you want the behaviour here to be locale-dependent
 * Add some tests which `setlocale()` to a Turkic or the Lithuanian locale and run tests there to check the behaviour

pwithnall (approved):
LGTM apart from the follow-up comment I left.

src/xb-machine.c
hughsie: I tried this just now; GMarkupParser seems to reject invalid UTF-8 before it ever gets to my callbacks with `G_MARKUP_ERROR_BAD_UTF8`

PR #79, opened 2021-01-22
Use g_str_match_string() for non-ASCII search queries

hughsie:
When searching for non-ASCII this slows down the query by ~ x4 and so this is
an interim fix before we tokenize as part of the silo build.

### DIFF
--- a/src/xb-silo.c
+++ b/src/xb-silo.c
@@ -1283,6 +1283,10 @@ xb_silo_machine_func_search_cb (XbMachine *self,
 				gpointer exec_data,
 				GError **error)
 {
+	XbSilo *silo = XB_SILO (user_data);
+	XbSiloPrivate *priv = GET_PRIVATE (silo);
+	const gchar *text;
+	const gchar *search;
 	XbOpcode *head1 = NULL;
 	XbOpcode *head2 = NULL;
 	g_auto(XbOpcode) op1 = XB_OPCODE_INIT ();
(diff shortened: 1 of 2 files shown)


src/xb-silo.c
```diff
@@ -1306,9 +1310,21 @@ xb_silo_machine_func_search_cb (XbMachine *self,
...
 		return FALSE;
 
+	/* this is going to be slow, but correct */
+	text = xb_opcode_get_str (&op2);
+	search = xb_opcode_get_str (&op1);
+	if (!g_str_is_ascii (text) || !g_str_is_ascii (search)) {
+		if (search == NULL || search[0] == '\0')
+			return xb_stack_push_bool (stack, FALSE, error);
+		if (priv->profile_flags & XB_SILO_PROFILE_FLAG_DEBUG) {
+			g_debug ("tokenization for [%s:%s] is will be slow!",
```
pwithnall, line 1320: Nitpick: ‘is will be’

pwithnall (commented):
LGTM. Is there some documentation somewhere about what the `~=` operator is actually defined to mean? I’m slightly worried that changing its behaviour like this might count as an API break, but it entirely depends how it’s defined.

src/xb-silo.c
hughsie: Fixed, ta.

hughsie:
> Is there some documentation somewhere about what the ~= operator is actually defined to mean

No, some XPath parsers connect it up to `search()` but it's not actually specified (IIRC) what search is supposed to match on.

pwithnall:
I guess if there are compatibility problems then we can always add a `search()` method with some arguments to control its behaviour (or several methods with different behaviours).

PR #80, opened 2021-01-22
Support adding tokens to the strtab at compile time

hughsie:
This allows us to normalize and tokenize at silo build time, which allows us
to perform `~=` queries on UTF-8 strings without using g_str_match_string()
in the fast path.

This allows us to search UTF-8 strings at the same speed as ASCII ones, at
the expense of a 15% larger xmlb blob.

Also add the --tokenize argument to xb-tool to allow building the silo in this
way.

    xb-tool compile test.xmlb fedora.xml.gz -v --tokenize name --tokenize summary
    xb-tool query test.xmlb "components/component/pkgname[text()='gimp']/../id"
    xb-tool query test.xmlb "components/component/name[text()~='manip']/../id"
    xb-tool query test.xmlb "components/component/name[text()~='изображ']/../id"


hughsie:
@pwithnall if that looks sane to you I can split it up in to a reviewable series rather than a megapatch.

pwithnall:
Review swap! I’ll look at this once you’ve reviewed https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/583. :)

hughsie:
Yes, I'll do that review next week, it's quite a bit of work. I'm on daddy duty for the next few days unfortunately.

ximion:
@pwithnall did most of that review already, there are just a few questions in that MR that really need your feedback (and once we have the answers, I can update the code with your feedback).

hughsie:
> pwithnall did most of that review already

Phew! Thanks @pwithnall, much appreciated. I've added some of my "_wisdom_" to a few places where it was requested :)

pwithnall:
I can review this later this week, once it’s split up. Presumably #82 needs to land first?

hughsie:
> I can review this later this week

Excellent. For now I just wanted a check this pre-tokenization is a good idea?

> Presumably #82 needs to land first?

Indeed.

pwithnall:
> Excellent. For now I just wanted a check this pre-tokenization is a good idea?

As far as I can tell from skim-reading the patch, yes.

PR #84, opened 2021-01-26
Allow adding search tokens to XbBuilderNode

hughsie:


src/xb-builder-node.c
```diff
@@ -297,6 +301,41 @@ xb_builder_node_parse_literal_text (XbBuilderNode *self, const gchar *text, gssi
 	return g_string_free (tmp, FALSE);
 }
 
+/**
+ * xb_builder_node_tokenize_text:
+ * @self: a #XbBuilderNode
+ *
+ * Tokenize text added with xb_builder_node_set_text().
```
pwithnall, line 308: Would be helpful to document what ‘tokenizing’ means.

Perhaps would be useful to add that as a paragraph or two (“nodes can be tokenized. This means blah blah and allows for blah blah with 2x better performance at a cost of 2x more storage”) in the section-level documentation for `XbBuilderNode`, and then link to it from here (and the other new method below).

src/xb-builder-node.c
```diff
@@ -297,6 +301,41 @@ xb_builder_node_parse_literal_text (XbBuilderNode *self, const gchar *text, gssi
...
+	XbBuilderNodePrivate *priv = GET_PRIVATE (self);
+	g_auto(GStrv) tokens = NULL;
+	g_auto(GStrv) ascii_tokens = NULL;
+
+	g_return_if_fail (XB_IS_BUILDER_NODE (self));
+
+	if (priv->text == NULL)
+		return;
+	tokens = g_str_tokenize_and_fold (priv->text, NULL, &ascii_tokens);
+	for (guint i = 0; tokens[i] != NULL; i++) {
```
pwithnall, line 324:
```suggestion
	for (gsize i = 0; tokens[i] != NULL; i++) {
```

(Array indices should be `gsize` as they could address all locations in memory if the array is big enough)

src/xb-builder-node.c
```diff
@@ -297,6 +301,41 @@ xb_builder_node_parse_literal_text (XbBuilderNode *self, const gchar *text, gssi
...
+		if (!xb_string_token_valid (tokens[i]))
+			continue;
+		xb_builder_node_add_token (self, tokens[i]);
```
pwithnall, line 327: If you expect roughly all the tokens to be valid, `xb_builder_node_add_token()` is going to be called a lot, and hence will need to expand the `GPtrArray` allocation quite a lot. Might be more performant to pre-allocate the right number of elements in the array.

src/xb-builder-node.c
```diff
@@ -297,6 +301,41 @@ xb_builder_node_parse_literal_text (XbBuilderNode *self, const gchar *text, gssi
...
+{
+	XbBuilderNodePrivate *priv = GET_PRIVATE (self);
+	g_auto(GStrv) tokens = NULL;
+	g_auto(GStrv) ascii_tokens = NULL;
+
+	g_return_if_fail (XB_IS_BUILDER_NODE (self));
+
+	if (priv->text == NULL)
+		return;
+	tokens = g_str_tokenize_and_fold (priv->text, NULL, &ascii_tokens);
```
pwithnall, line 323: Might want to hook `translit_locale` up to an argument in `xb_builder_node_tokenize_text()`, as I imagine the caller might have information about the locale of the source string in some cases.

src/xb-builder-node.c
pwithnall: It’s also going to call `g_strdup()` on all of the tokens, which is unnecessary given that you could steal them out of the `GStrv`, in this case anyway.

src/xb-builder-node.c
```diff
@@ -1107,3 +1151,43 @@ xb_builder_node_export (XbBuilderNode *self, XbNodeExportFlags flags, GError **e
...
+	if (priv->tokens == NULL)
+		priv->tokens = g_ptr_array_new_with_free_func (g_free);
+	g_ptr_array_add (priv->tokens, g_strdup (token));
+}
+
+/**
+ * xb_builder_node_get_tokens:
+ * @self: a #XbBuilderNode
+ *
+ * Gets the tokens of the builder node.
```
pwithnall, line 1181: Definitely needs some docs to explain what ‘tokens’ are. I suggest linking to the top-level documentation section on it.

src/xb-builder-node.h
```diff
@@ -34,6 +34,7 @@ struct _XbBuilderNodeClass {
  * @XB_BUILDER_NODE_FLAG_LITERAL_TEXT:		Assume the node CDATA is already valid
  * @XB_BUILDER_NODE_FLAG_HAS_TEXT:		If the node has leading text
  * @XB_BUILDER_NODE_FLAG_HAS_TAIL:		If the node has trailing text
+ * @XB_BUILDER_NODE_FLAG_TOKENIZE_TEXT:		Tokenize and fold text to ASCII
```
pwithnall, line 37:
```suggestion
 * @XB_BUILDER_NODE_FLAG_TOKENIZE_TEXT:		Tokenize and fold text to ASCII (Since: 0.3.1)
```

src/xb-string.c
```diff
@@ -178,6 +178,26 @@ xb_string_search (const gchar *text, const gchar *search)
 	return FALSE;
 }
 
+/**
+ * xb_string_token_valid: (skip)
+ * @text: The potential token
+ *
+ * Checks if the token is valid.
```
pwithnall, line 185: Could do with some docs to say what ‘valid’ means, and what the purpose of calling this function is (i.e. something like “to eliminate tokens which are not useful for search matching”)

src/xb-builder-node.c
hughsie: hmm, maybe the galaxy brain thing to do would be to use the `xml:lang` attribute -- this makes it magically just work with AppStream...

src/xb-builder-node.c
hughsie: I've added something.

src/xb-builder-node.c
hughsie: > Might be more performant to pre-allocate the right number of elements in the array

I think I've done something like that; no observable difference in compile speed but I guess it all adds up.

src/xb-builder-node.c
hughsie: Yes, that works.

src/xb-builder-node.c
hughsie: Added I think

src/xb-string.c
hughsie: Added I think.

PR #85, opened 2021-01-26
Wip/hughsie/xb opcode has flag

hughsie:


src/xb-opcode.c
```diff
@@ -176,9 +214,7 @@ xb_opcode_cmp_val (XbOpcode *self)
 inline gboolean
 xb_opcode_cmp_str (XbOpcode *self)
 {
-	return self->kind == XB_OPCODE_KIND_TEXT ||
-		self->kind == XB_OPCODE_KIND_BOUND_TEXT ||
-		self->kind == XB_OPCODE_KIND_INDEXED_TEXT;
+	return (self->kind & XB_OPCODE_FLAG_TEXT) > 0;
```
pwithnall, line 217: Couldn’t this be a call to `xb_opcode_has_flag()`?

hughsie: It can indeed, fixed.

pwithnall (approved):
LGTM

PR #86, opened 2021-01-30
Wip/hughsie/xb tool tokenize

hughsie:


src/xb-tool.c
```diff
@@ -456,6 +481,8 @@ main (int argc, char *argv[])
 			"Return only when the silo is no longer valid", NULL },
 		{ "profile", 'p', 0, G_OPTION_ARG_NONE, &priv->profile,
 			"Show profiling information", NULL },
+		{ "tokenize", 'p', 0, G_OPTION_ARG_STRING_ARRAY, &priv->tokenize,
```
pwithnall, line 484: You can’t re-use the same `'p'` short name for the option here; it’s already used by `--profile`. Might also help to give an example of what values to provide to `--tokenize`.

src/xb-builder-node.c
```diff
@@ -1219,3 +1222,26 @@ xb_builder_node_get_tokens (XbBuilderNode *self)
...
+	g_return_if_fail (tail_idx != XB_SILO_UNSET);
+
+	if (priv->token_idxs == NULL)
+		priv->token_idxs = g_array_new (FALSE, FALSE, sizeof(guint32));
+	g_array_append_val (priv->token_idxs, tail_idx);
+}
+
+/* private */
+GArray *
+xb_builder_node_get_token_idxs (XbBuilderNode *self)
```
pwithnall, line 1242: Might help to add a comment by this function saying that its return value is nullable.

src/xb-opcode-private.h
```diff
@@ -12,14 +12,17 @@
 
 G_BEGIN_DECLS
 
+#define XB_OPCODE_TOKEN_MAX		32
```
pwithnall, line 15: Could do with a comment mentioning what it’s for and why the value was chosen as 32.

src/xb-opcode.c
```diff
@@ -257,6 +268,22 @@ xb_opcode_get_str (XbOpcode *self)
 	return self->ptr;
 }
 
+/**
+ * xb_opcode_get_tokens:
+ * @self: a #XbOpcode
+ *
+ * Gets the tokenized string stored on the opcode.
+ *
+ * Returns: a #GStrv, or %NULL if unset
```
pwithnall, line 277: Add `(nullable)`

src/xb-string.c
```diff
@@ -178,6 +178,32 @@ xb_string_search (const gchar *text, const gchar *search)
...
 
+/**
+ * xb_string_search: (skip)
+ * @text: The source string
+ * @search: The text to search for
+ *
+ * Searches for a fuzzy search match, ignoring search matches that are not at
+ * the start of the token.
+ *
+ * Returns: %TRUE if the string @search is contained in @text.
```
pwithnall, line 189: Needs a `Since:` line

src/xb-string.c
```diff
@@ -178,6 +178,32 @@ xb_string_search (const gchar *text, const gchar *search)
 	return FALSE;
 }
 
+/**
+ * xb_string_search: (skip)
+ * @text: The source string
+ * @search: The text to search for
+ *
+ * Searches for a fuzzy search match, ignoring search matches that are not at
+ * the start of the token.
```
pwithnall, line 187: What’s a ‘fuzzy’ search match? Perhaps some examples would help.

src/xb-string.c
```diff
@@ -178,6 +178,32 @@ xb_string_search (const gchar *text, const gchar *search)
 	return FALSE;
 }
 
+/**
+ * xb_string_search: (skip)
```
pwithnall, line 182: `searchv` (the `v` is missing)

src/xb-string.c
```diff
@@ -178,6 +178,32 @@ xb_string_search (const gchar *text, const gchar *search)
...
+ * @text: The source string
+ * @search: The text to search for
```
pwithnall, line 184: This should probably say that `search` is actually a (`NULL` terminated?) array of somethings

src/xb-string.c
```diff
@@ -178,6 +178,32 @@ xb_string_search (const gchar *text, const gchar *search)
 	return FALSE;
 }
 
+/**
+ * xb_string_search: (skip)
+ * @text: The source string
```
pwithnall, line 183: This should probably say that `text` is actually a (`NULL` terminated?) array of somethings

src/xb-tool.c
hughsie: Ahh, typo, fixed. I've also added an example, thanks.

src/xb-builder-node.c
hughsie: Fixed, ta.

src/xb-opcode-private.h
hughsie: Fixed.

src/xb-opcode.c
hughsie: I don't think it is -- it's a static array of size `XB_OPCODE_TOKEN_MAX+1` i.e. always allocated.

src/xb-string.c
hughsie: Fixed, thanks.

src/xb-string.c
hughsie: Added

src/xb-string.c
hughsie: Ta.

src/xb-string.c
hughsie: I've fixed up that and given an example.

src/xb-string.c
hughsie: Fixed.

src/xb-opcode.c
pwithnall: Ooops, I got mixed up with the `token_idxs` array, which is lazily allocated in `xb-builder-node.c`. Sorry!

pwithnall: But the “or `%NULL` if unset” does need dropping in that case

src/xb-opcode.c
```diff
@@ -413,6 +440,17 @@ xb_opcode_set_val (XbOpcode *self, guint32 val)
...
 }
 
+/* private */
+void
+xb_opcode_set_token (XbOpcode *self, guint idx, const gchar *val)
+{
+	g_return_if_fail (val != NULL);
+	g_return_if_fail (val[0] != '\0');
+	g_return_if_fail (idx < XB_OPCODE_TOKEN_MAX);
+	self->tokens[idx] = val;
```
pwithnall, line 450: `self->tokens` is treated as a `NULL`-terminated array elsewhere in this file, so this function should ensure the `NULL` terminator is correct. That can’t happen if, for example, a caller calls `set_token(0, "foo")` then `set_token(15, "disjoint")`. It doesn’t happen now, but could happen later on if the code is refactored.

My suggestion would be to rename this `xb_opcode_append_token()` and drop the `idx` argument. Return a `gboolean` which is `TRUE` for success, or `FALSE` if the tokens array is full. That would simplify error handling in the callers (they would no longer have to check `XB_OPCODE_TOKEN_MAX` themselves) and might mean you could bring `XB_OPCODE_TOKEN_MAX` into this C file as a private implementation detail (although I haven’t worked through whether that’s possible).

src/xb-silo.c
```diff
@@ -1077,6 +1077,47 @@ xb_silo_machine_fixup_attr_exists_cb (XbMachine *self,
 	return TRUE;
 }
 
+static gboolean
+xb_silo_machine_fixup_attr_search_tokn_cb (XbMachine *self,
```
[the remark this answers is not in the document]
hughsie: Not really :) Fixed.

src/xb-opcode.c
hughsie: Agree, fixed.

src/xb-opcode.c
hughsie: > My suggestion would be to rename this xb_opcode_append_token() and drop the idx argument

Would we have to scan all the elements before adding each one? Doesn't that make each append check the NULLness a factorial number of times? Or do we make _XbOpcode slightly larger and store a `tokens_len` kind of thing?

pwithnall: Probably add a `tokens_len`

hughsie: Done!

pwithnall:
LGTM.

PR #88, opened 2021-03-23
Improve performance of XbStack

pwithnall:
Avoid heap-allocating it when it’s being used to run an `XbMachine`, as that happens 3 million times when starting gnome-software.

This reduces the time it takes to show the main window of gnome-software by 0.1s (from 1.1s to 1.0s). It reduces the number of `memset()` calls measured by callgrind from 27.4 million to 4.8 million.

No public API changes, so this should be OK to backport to stable distros if desired.

### DIFF
--- a/src/xb-stack-private.h
+++ b/src/xb-stack-private.h
@@ -8,9 +8,49 @@
 
 #include <glib-object.h>
 
+#include "xb-opcode-private.h"
 #include "xb-stack.h"
 
 G_BEGIN_DECLS
+
+/* Members of this struct should not be accessed directly — use the accessor
+ * functions below instead. */
+struct _XbStack {
+	/*< private >*/
+	gint		 ref;
+	gboolean	 stack_allocated;	/* whether this XbStack was allocated with alloca() */
+	guint		 pos;	/* index of the next unused entry in .opcodes */
+	guint		 max_size;
+	XbOpcode	 opcodes[];	/* allocated as part of XbStack */
+};
+
+/**
+ * xb_stack_new_inline:
+ * @max_stack_size: maximum size of the stack
+ *
+ * Creates a stack for the XbMachine request. Only #XbOpcodes can be pushed and
+ * popped from the stack.
+ *
+ * The stack will be allocated on the current C stack frame, so @max_stack_size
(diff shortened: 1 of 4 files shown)


src/xb-stack-private.h
```diff
@@ -8,10 +8,50 @@
...
+ * The stack will be allocated on the current C stack frame, so @max_stack_size
+ * should be chosen to not overflow the C process’ stack.
+ *
+ * Returns: (transfer full): a #XbStack
+ *
+ * Since: 0.3.1
+ **/
+#define xb_stack_new_inline(max_stack_size) \
+(G_GNUC_EXTENSION ({ \
+	/* This function has to be static inline so we can use g_alloca(), which \
```
hughsie, line 43: I'm a bit unclear on why it needs to be static inline. Got any background to make me sound like a Unix newbie?

pwithnall: oops, that comment is out of date — it’s a macro, rather than `static inline`. Making it `static inline` doesn’t work.

It needs to be a macro because allocations made with `alloca()` live as long as the stack frame they’re allocated in, so the only way to abstract them is to move them into a macro.

hughsie: I was going to say "why can't we just do XbStack stack = XB_STACK_INIT" but then that's basically what this is. Perhaps we could make it more like XB_QUERY_CONTEXT_INIT in use rather than pretending it's an action function.

pwithnall: I’m not sure that would make things clearer, since `XB_QUERY_CONTEXT_INIT` follows the pattern for stack-allocated, non-pointer variables. `alloca()` is a weird one in that it creates a stack-allocated, pointer variable.

hughsie: Ya, indeed. I'm not sure why I'm getting so uptight, this is just internal non-exported API.

hughsie (approved):

src/xb-stack-private.h
pwithnall: It’s good to get things right even for internal API, thanks for checking! :slightly_smiling_face:

PR #89, opened 2021-04-30
trivial: Fix a win32 failure when g_content_type_guess() returns bad values

hughsie:


src/xb-common.c
```diff
@@ -96,3 +97,44 @@ xb_content_type_guess (const gchar *filename, const guchar *buf, gsize bufsz)
...
+#endif
+
+	g_return_val_if_fail (G_IS_FILE (file), FALSE);
+	g_return_val_if_fail (buf != NULL, FALSE);
+	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+#if defined(_WIN32) && GLIB_CHECK_VERSION(2,66,0)
+	return g_file_set_contents_full (fn, (const gchar *) buf, (gssize) bufsz,
+					 G_FILE_SET_CONTENTS_NONE, 0666, error);
```
superm1, line 131: so what about WIN32 and older glib though?  Does that mean it will fail there?  Should it be an error?

hughsie: Yes, you're right. Fixed!

superm1 (approved):

PR #92, opened 2021-05-21
Switch from Travis to Github actions

hughsie:


hughsie:
@Asiderr aaaaaand another please.

Asiderr (changes requested):
Does not work on FreeBSD 12.2. Again needed change https://github.com/hughsie/libxmlb/blob/master/meson.build#L152:
```
+ python3 = python.find_installation()
- python3 = python.find_installation('python3')
```

hughsie:
Doh, thanks.

Asiderr (approved):

PR #94, opened 2021-07-10
Add support for LZMA decompression

hughsie:
Compressing XML files with xz compared to gz produces files that are 30% smaller.


ximion:
Is this for some embedded usecases where space is extremely precious but CPU power not so much?
(I actually played a bit with XZ, GZip and ZStandard on XML in the past, and while XZ was slowest on my machine by a few dozen milliseconds, it actually barely mattered which algorithm was used unless the XML was huge (but gzip was always faster than loading raw data) - on a slower machine though it most certainly will have a huge impact though)
This is just a curiosity question, especially since ZStandard would also reduce size a bit but not sacrifice as much CPU time for decompression.

hughsie:
> Is this for some embedded usecases where space is extremely precious

Not storage per-se. There are ~10 million downloading a 616Kb file globally per day on the LVFS. We could **save** 1.5TB of bandwidth per day switching to xz! From a fwupd point of view, we would download the .xz, decompress it, parse it, and then we'd actually save it in /usr/lib/fwupd as .xml.gz as like you say it's quickest to parse for next time we regenerate the silo. I also went with LZMA rather than ZStandard as I already need xz-devel in fwupd for the EFI decoding, and the differences between lmza and zstd were much much less significant than the differences between gzip and lzma.

superm1 (approved):

ximion:
> Not storage per-se. There are ~10 million downloads of a 616Kb file globally per day on the LVFS. We could save 1.5TB of bandwidth per day switching to xz! 

Makes a lot of sense! :-) At a certain scale, tine improvements are huge!

> [...] the differences between lmza and zstd were much much less significant than the differences between gzip and lzma.

Matches my experience with compressed AppStream XML as well :-)

PR #96, opened 2021-09-20
Fix GMainContext issues with silo signals and file monitoring

pwithnall:
See the commit messages for details.

This is not required to fix the immediate issue in gnome-software (a workaround was committed in https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1422), but does tidy things up and make the behaviour more consistent.

Code which uses libxmlb should check what `GMainContext` is the thread default when calling `xb_builder_new()`, `xb_silo_new()`, `xb_builder_ensure()`, `xb_builder_compile()` or `xb_silo_watch_file()`.

### DIFF
--- a/src/xb-silo.c
+++ b/src/xb-silo.c
@@ -1565,7 +1565,7 @@ xb_silo_class_init (XbSiloClass *klass)
 				     G_PARAM_STATIC_NAME);
 
 	/**
-	 * XbSilo:allow-cancel:
+	 * XbSilo:valid:
 	 */
 	obj_props[PROP_VALID] =
 		g_param_spec_boolean ("valid", NULL, NULL, TRUE,
--- a/src/xb-silo.c
+++ b/src/xb-silo.c
@@ -4,6 +4,24 @@
  * SPDX-License-Identifier: LGPL-2.1+
  */
 
+/**
+ * SECTION:xb-silo
+ * @title: XbSilo
+ * @include: xmlb.h
+ * @stability: Stable
+ * @short_description: A read-only store of parsed XML data
+ *
+ * #XbSilo provides read-only access and querying of a previously parsed blob
+ * of XML data.
+ *
+ * All signal emissions from #XbSilo (currently only #GObject::notify emissions)
+ * will happen in the #GMainContext which is the thread default when the #XbSilo
+ * is constructed.
+ *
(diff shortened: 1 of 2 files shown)


src/xb-silo.c
```diff
@@ -58,13 +58,13 @@ typedef struct {
...
-	PROP_GUID,
+typedef enum {
+	PROP_GUID = 1,
 	PROP_VALID,
 	PROP_ENABLE_NODE_CACHE,
-	PROP_LAST
-};
+} XbSiloProperty;
+
+static GParamSpec *obj_props[PROP_ENABLE_NODE_CACHE + 1] = { NULL, };
```
hughsie, line 87: not PROP_LAST? That seemed a nice thing.

src/xb-silo.c
```diff
@@ -833,24 +913,52 @@ xb_silo_watch_file (XbSilo *self,
...
+static gboolean
+watch_file_cb (gpointer user_data)
+{
+	g_autoptr(WatchFileData) data = g_steal_pointer (&user_data);
+	XbSilo *self = data->silo;
+	GFile *file = data->file;
+	XbSiloFileMonitorItem *item;
+	XbSiloPrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GFileMonitor) file_monitor = NULL;
+	g_autoptr(GError) local_error = NULL;
```
hughsie, line 948: I think elsewhere we use error_local.

src/xb-silo.c
pwithnall: No, because `PROP_LAST` is not really a member of the enum, it’s metadata about the enum. Unfortunately there’s no way in C to do something like `number_of_elements_in (XbSiloProperty)`. `PROP_LAST` is a hack to achieve that, but if used with `-Wswitch-enum` it means that every `switch` statement which handles values of that enum type will now have to have a `case PROP_LAST: g_assert_not_reached()` case, which is painful.

Using `PROP_ENABLE_NODE_CACHE + 1` in the array definition here is the least-worst solution I know. The compiler should catch if we fail to update the definition in future, since we’d be statically assigning to an element off the end of the array in `class_init` if so.

src/xb-silo.c
pwithnall: Changed.

hughsie (approved):

PR #98, opened 2021-11-12
Use -Dcli=false to reduce the install size

hughsie:
We don't need or want this debugging tool on embedded targets.


hughsie:
@rossburton FYI

rossburton:
Is 'cli' the right name for this, considering there are other commandline tools that still get installed?

hughsie:
> there are other commandline tools

which ones? I think that's the only binary we install from libxmlb....

hughsie:
If there's a more standard name "e.g. `-Ddebug` perhaps? I'm all ears :)

rossburton:
Sorry, for some reason I thought this was another package. 'cli' seems fine.

superm1 (approved):

PR #99, opened 2021-11-24
Add a flag to require the XbBuilderSource to have no siblings

hughsie:
This is required for programs like fwupd to reject invalid (appended)
metainfo.xml files.

### DIFF
--- a/src/xb-builder.c
+++ b/src/xb-builder.c
@@ -260,6 +260,17 @@ xb_builder_compile_source (XbBuilderCompileHelper *helper,
 	if (!xb_builder_source_fixup (source, root_tmp, error))
 		return FALSE;
 
+	/* a single root with no siblings was required */
+	if (helper->compile_flags & XB_BUILDER_COMPILE_FLAG_SINGLE_ROOT) {
+		if (xb_builder_node_get_children (root_tmp)->len > 1) {
+			g_set_error_literal (error,
+					     G_IO_ERROR,
+					     G_IO_ERROR_INVALID_DATA,
+					     "A root node without siblings was required");
+			return FALSE;
+		}
+	}
+
 	/* this is something we can query with later */
 	info = xb_builder_source_get_info (source);
 	if (info != NULL) {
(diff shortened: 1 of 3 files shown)


src/xb-self-test.c
```diff
@@ -2641,6 +2659,7 @@ main (int argc, char **argv)
 	g_test_add_func ("/libxmlb/xpath-node", xb_xpath_node_func);
 	g_test_add_func ("/libxmlb/xpath-parent-subnode", xb_xpath_parent_subnode_func);
 	g_test_add_func ("/libxmlb/multiple-roots", xb_builder_multiple_roots_func);
+	g_test_add_func ("/libxmlb/require-root", xb_builder_require_root_func);
```
pwithnall, line 2662:
```suggestion
	g_test_add_func ("/libxmlb/single-root", xb_builder_single_root_func);
```

aiui that name would better reflect what’s being tested

src/xb-builder.c
```diff
@@ -260,6 +260,17 @@ xb_builder_compile_source (XbBuilderCompileHelper *helper,
 	if (!xb_builder_source_fixup (source, root_tmp, error))
 		return FALSE;
 
+	/* a single root with no siblings was required */
+	if (helper->compile_flags & XB_BUILDER_COMPILE_FLAG_SINGLE_ROOT) {
+		if (xb_builder_node_get_children(root_tmp)->len > 1) {
```
pwithnall, line 265:
```suggestion
		if (xb_builder_node_get_children (root_tmp)->len > 1) {
```

src/xb-builder.h
```diff
@@ -39,6 +39,7 @@ struct _XbBuilderClass {
  * @XB_BUILDER_COMPILE_FLAG_SINGLE_LANG:	Only store a single language
  * @XB_BUILDER_COMPILE_FLAG_WATCH_BLOB:		Watch the XMLB file for changes
  * @XB_BUILDER_COMPILE_FLAG_IGNORE_GUID:	Ignore the cache GUID value
+ * @XB_BUILDER_COMPILE_FLAG_SINGLE_ROOT:	Require a root node without siblings
```
pwithnall, line 42: Perhaps “Require at most one root node”? Otherwise it’s ambiguous whether this results in an error if the document has no root nodes.

pwithnall (changes requested):

src/xb-builder.h
hughsie: Agreed, thanks. Fixed.

pwithnall (commented):
LGTM!

hughsie:
@pwithnall thanks for the super-speedy review dude.

PR #100, opened 2021-11-24
Add simple iterator for XbNode attributes

ximion:
Hi!
This adds the simplest iterator possible to iterate over all XbNode attributes (can only initialize and go forward). With having something like this, I can convert an XbNode into a libxml2 node using some really simple test code:
```c
static gboolean
as_transmogrify_xbnode_to_xmlnode (XbNode *xbn, xmlNode *lxn)
{
	XbNodeIter attr_iter;
	const gchar *attr_name;
	const gchar *attr_value;
	g_autoptr(GPtrArray) children = NULL;
	g_return_val_if_fail (XB_IS_NODE (xbn), FALSE);

	xmlNodeSetName (lxn,
			(xmlChar*) xb_node_get_element (xbn));
	xmlNodeAddContent (lxn,
			   (xmlChar*) xb_node_get_text (xbn));

	xb_node_attr_iter_init (&attr_iter, xbn);
	while (xb_node_attr_iter_next (&attr_iter, &attr_name, &attr_value))
		xmlNewProp (lxn, (xmlChar*) attr_name, (xmlChar*) attr_value);

	/* all children */
	children = xb_node_get_children (xbn);
	for (guint i = 0; i < children->len; i++) {
		XbNode *c = XB_NODE (g_ptr_array_index (children, i));
		xmlNode *lc = xmlNewChild (lxn, NULL, (xmlChar*) "", NULL);
		if (!as_transmogrify_xbnode_to_xmlnode (c, lc))
			return FALSE;
	}

	return TRUE;
}
```
In AppStream, this reduces the time to show results from my test query by 70-80msec compared to serializing the XbNode to XML and loading it with libxml2 again, which is rather nice!

I intend to maybe extend this and add an iterator for XbNode children as well, reusing the new `XbNodeIter` struct for that as well, to save a GPtrArray allocation and looping through the array twice.

Thanks for considering!

### DIFF
--- a/src/xb-node.c
+++ b/src/xb-node.c
@@ -24,6 +24,32 @@ G_DEFINE_TYPE_WITH_PRIVATE (XbNode, xb_node, G_TYPE_OBJECT)
 #define GET_PRIVATE(o) (xb_node_get_instance_private (o))
 
 /**
+ * XbNodeAttrIter:
+ *
+ * A #XbNodeAttrIter structure represents an iterator that can be used
+ * to iterate over the attributes of a #XbNode. #XbNodeAttrIter
+ * structures are typically allocated on the stack and then initialized
+ * with xb_node_attr_iter_init().
+ *
+ * The iteration order of a #XbNodeAttrIter is not defined.
+ *
+ * Since: 0.3.4
+ */
+
+typedef struct
+{
+  XbNode 	*node;
+  guint8	 position;
+  gpointer	 dummy3;
+  gpointer	 dummy4;
+  gpointer	 dummy5;
+  gpointer	 dummy6;
+} RealAttrIter;
+
+G_STATIC_ASSERT (sizeof (XbNodeAttrIter) == sizeof (RealAttrIter));
+G_STATIC_ASSERT (G_ALIGNOF (XbNodeAttrIter) >= G_ALIGNOF (RealAttrIter));
+
+/**
  * xb_node_get_data:
  * @self: a #XbNode
  * @key: a string key, e.g. `fwupd::RemoteId`
@@ -363,6 +389,82 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
 	if (g_str_has_prefix (tmp, "0x"))
 		return g_ascii_strtoull (tmp + 2, NULL, 16);
 	return g_ascii_strtoull (tmp, NULL, 10);
+}
+
+/**
+ * xb_node_attr_iter_init:
+ * @iter: an uninitialized #XbNodeAttrIter
+ * @self: a #XbNode
+ *
+ * Initializes a name/value pair iterator for the node attributes
+ * and associates it with @self.
+ * The #XbNodeAttrIter structure is typically allocated on the stack
+ * and does not need to be freed explicitly.
+ *
+ * Since: 0.3.4
+ */
+void
+xb_node_attr_iter_init (XbNodeAttrIter *iter, XbNode *self)
+{
+	XbNodePrivate *priv = GET_PRIVATE (self);
+	RealAttrIter *ri = (RealAttrIter *) iter;
+
+	g_return_if_fail (iter != NULL);
+	g_return_if_fail (XB_IS_NODE (self));
+
+	ri->node = self;
+	ri->position = xb_silo_node_get_attr_count (priv->sn);
+}
+
+/**
+ * xb_node_attr_iter_next:
+ * @iter: an initialized #XbNodeAttrIter
+ * @name: (out) (optional) (not nullable): Destination of the returned attribute name
+ * @value: (out) (optional) (not nullable): Destination of the returned attribute value
+ *
+ * Returns the current attribute name and value and advances the iterator.
+ * Example:
+ * |[<!-- language="C" -->
+ * XbNodeAttrIter iter;
+ * const gchar *attr_name, *attr_value;
+ *
+ * xb_node_attr_iter_init (&iter, node);
+ * while (xb_node_attr_iter_next (&iter, &attr_name, &attr_value)) {
+ *     // use attr_name and attr_value; no need to free them
+ * }
+ * ]|
+ *
+ * Returns: %TRUE if there are more attributes.
+ *
+ * Since: 0.3.4
+ */
+gboolean
+xb_node_attr_iter_next (XbNodeAttrIter *iter, const gchar **name, const gchar **value)
+{
+	XbSiloNodeAttr *a;
+	XbNodePrivate *priv;
+	RealAttrIter *ri = (RealAttrIter *) iter;
+
+	g_return_val_if_fail (iter != NULL, FALSE);
+	priv = GET_PRIVATE (ri->node);
+
+	/* check if the iteration was finished */
+	if (ri->position == 0) {
+		if (name != NULL)
+			*name = NULL;
+		if (value != NULL)
+			*value = NULL;
+		return FALSE;
+	}
+
+	ri->position--;
+	a = xb_silo_node_get_attr (priv->sn, ri->position);
+	if (name != NULL)
+		*name = xb_silo_from_strtab (priv->silo, a->attr_name);
+	if (value != NULL)
+		*value = xb_silo_from_strtab (priv->silo, a->attr_value);
+
+	return TRUE;
 }
 
 /**
--- a/src/xb-node.h
+++ b/src/xb-node.h
@@ -49,6 +49,17 @@ typedef enum {
 	XB_NODE_EXPORT_FLAG_LAST
 } XbNodeExportFlags;
 
+typedef struct
+{
+  /*< private >*/
+  gpointer	dummy1;
+  guint8	dummy2;
+  gpointer	dummy3;
+  gpointer	dummy4;
+  gpointer	dummy5;
+  gpointer	dummy6;
+} XbNodeAttrIter;
+
 typedef gboolean (*XbNodeTransmogrifyFunc)	(XbNode		*self,
 						 gpointer	 user_data);
 gboolean	 xb_node_transmogrify		(XbNode		*self,
@@ -80,5 +91,10 @@ guint64		 xb_node_get_attr_as_uint	(XbNode		*self,
 						 const gchar	*name);
 guint		 xb_node_get_depth		(XbNode		*self);
 
+void		 xb_node_attr_iter_init		(XbNodeAttrIter	*iter,
+						 XbNode		*self);
+gboolean	 xb_node_attr_iter_next		(XbNodeAttrIter	*iter,
+						 const gchar	**name,
+						 const gchar	**value);
 
 G_END_DECLS
--- a/src/libxmlb.map
+++ b/src/libxmlb.map
@@ -260,3 +260,10 @@ LIBXMLB_0.3.1 {
     xb_builder_node_tokenize_text;
   local: *;
 } LIBXMLB_0.3.0;
+
+LIBXMLB_0.3.4 {
+  global:
+    xb_node_attr_iter_init;
+    xb_node_attr_iter_next;
+  local: *;
+} LIBXMLB_0.3.1;


ximion:
I just added those sanity checks I was talking about before - they should be really cheap, and other XbNode functions also verify their parameters at least a little bit, so the new ones should do that too.

src/xb-node.c
```diff
@@ -18,6 +18,8 @@
 typedef struct {
 	XbSilo			*silo;
 	XbSiloNode		*sn;
+
+	guint8			attr_iter_pos;
```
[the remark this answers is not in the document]
ximion: That would make the code a bit more complex, but sure, I can add an `XbNodeAttrIterator` :-)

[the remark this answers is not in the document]
hughsie: XbNodeAttrIter perhaps?

src/xb-node.c
```diff
@@ -18,11 +18,37 @@
...
```
hughsie, line 22: this can go now, right?

ximion:
We now have a GHashTableIter-style iterator for this, so this should be threadsafe now. I called the iterator `XbNodeIter`, because I think it may be legit to repurpose the same iterator for iterating over the node's children as well (recycling the iter struct, but using two new functions to do the iteration).

hughsie:
This like what I imaged, thanks.

src/xb-node.c
```diff
@@ -23,6 +23,30 @@ typedef struct {
...
 
+/**
+ * XbNodeIter:
+ *
+ * A XbNodeIter structure represents an iterator that can be used
+ * to iterate over the attributes of a #XbNode. XbNodeIter
+ * structures are typically allocated on the stack and then initialized
+ * with %xb_node_attr_iter_init().
+ *
+ * The iteration order of a #XbNodeIter is not defined.
```
pwithnall, line 34:
```suggestion
 * The iteration order of a #XbNodeIter is not defined.
 *
 * Since: 0.3.4
```

src/xb-node.c
```diff
@@ -23,6 +23,30 @@ typedef struct {
 G_DEFINE_TYPE_WITH_PRIVATE (XbNode, xb_node, G_TYPE_OBJECT)
 #define GET_PRIVATE(o) (xb_node_get_instance_private (o))
 
+/**
+ * XbNodeIter:
+ *
+ * A XbNodeIter structure represents an iterator that can be used
+ * to iterate over the attributes of a #XbNode. XbNodeIter
```
pwithnall, line 30:
```suggestion
 * to iterate over the attributes of a #XbNode. #XbNodeIter
```

src/xb-node.c
```diff
@@ -23,6 +23,30 @@ typedef struct {
...
+ * structures are typically allocated on the stack and then initialized
+ * with %xb_node_attr_iter_init().
```
pwithnall, line 32:
```suggestion
 * with xb_node_attr_iter_init().
```

src/xb-node.c
```diff
@@ -23,6 +23,30 @@ typedef struct {
...
+ *
+ * The iteration order of a #XbNodeIter is not defined.
+ */
+
+typedef struct
+{
+  XbNode 	*node;
+  guint8	position;
```
pwithnall, line 40: Alignment is off for `position` relative to `node`. Either don’t align any of the members’ names, or align all of them.

src/xb-node.c
```diff
@@ -23,6 +23,30 @@ typedef struct {
...
+  gpointer	dummy3;
+  gpointer	dummy4;
+  gpointer	dummy5;
+  gpointer	dummy6;
+} RealIter;
```
pwithnall, line 45: Might want to call this `RealNodeIter` in case another kind of iter is added to this file in future.

src/xb-node.c
```diff
@@ -365,6 +389,70 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
...
+/**
+ * xb_node_attr_iter_init:
+ * @iter: an uninitialized #XbNodeIter
+ * @self: a #XbNode
+ *
+ * Initializes a name/value pair iterator for the node attributes
+ * and associates it with @self.
+ *
+ * Since: 0.3.4
+ **/
```
pwithnall, line 401:
```suggestion
 */
```

Finishing a doc comment with a double-star is long deprecated.

src/xb-node.c
```diff
@@ -365,6 +389,70 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
 	return g_ascii_strtoull (tmp, NULL, 10);
 }
 
+/**
+ * xb_node_attr_iter_init:
+ * @iter: an uninitialized #XbNodeIter
+ * @self: a #XbNode
+ *
+ * Initializes a name/value pair iterator for the node attributes
+ * and associates it with @self.
```
pwithnall, line 398: Might want to mention whether any `clear()` or `free()` function eventually needs to be called on a `XbNodeIter` initialised this way; or explicitly state that’s not necessary.

src/xb-node.c
```diff
@@ -365,6 +389,70 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
...
+ * @iter: an initialized #XbNodeIter
+ * @name: (out): Destination of the returned attribute name
+ * @value: (out): Destination of the returned attribute value
+ *
+ * Returns the current attribute name and value and advances the iterator.
+ *
+ * Returns: %TRUE if there are more attributes.
+ *
+ * Since: 0.3.4
+ **/
```
pwithnall, line 426:
```suggestion
 */
```

src/xb-node.c
```diff
@@ -365,6 +389,70 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
...
+	ri->position = xb_silo_node_get_attr_count (priv->sn);
+}
+
+/**
+ * xb_node_attr_iter_next:
+ * @iter: an initialized #XbNodeIter
+ * @name: (out): Destination of the returned attribute name
+ * @value: (out): Destination of the returned attribute value
+ *
+ * Returns the current attribute name and value and advances the iterator.
```
pwithnall, line 421: It would be helpful to include a brief code example of how to do a loop with `xb_node_attr_iter_{init,next}()`, which people can just copy and adapt into their code.

e.g.
```
 * |[
 * XbNodeAttrIter iter;
 * const gchar *attr_name, *attr_value;
 *
 * xb_node_attr_iter_init (&iter, node);
 * while (xb_node_attr_iter_next (&iter, &attr_name, &attr_value))
 *   {
 *     /* use attr_name and attr_value; no need to free them */
 *   }
 * ]|
```

src/xb-node.c
```diff
@@ -365,6 +389,70 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
...
+
+	ri->node = self;
+	ri->position = xb_silo_node_get_attr_count (priv->sn);
+}
+
+/**
+ * xb_node_attr_iter_next:
+ * @iter: an initialized #XbNodeIter
+ * @name: (out): Destination of the returned attribute name
+ * @value: (out): Destination of the returned attribute value
```
pwithnall, line 419:
```suggestion
 * @name: (out) (optional) (not nullable): Destination of the returned attribute name
 * @value: (out) (optional) (not nullable): Destination of the returned attribute value
```

src/xb-node.c
```diff
@@ -365,6 +389,70 @@ xb_node_get_attr_as_uint (XbNode *self, const gchar *name)
...
+xb_node_attr_iter_next (XbNodeIter *iter, const gchar **name, const gchar **value)
+{
+	XbSiloNodeAttr *a;
+	XbNodePrivate *priv;
+	RealIter *ri = (RealIter *) iter;
+
+	g_return_val_if_fail (iter != NULL, FALSE);
+	priv = GET_PRIVATE (ri->node);
+
+	/* check if the iteration was started */
```
pwithnall, line 437:
```suggestion
	/* check if the iteration was finished */
```

Maybe? Not sure if I understand this comment properly.

src/xb-node.h
```diff
@@ -49,6 +49,19 @@ typedef enum {
...
+struct _XbNodeIter
+{
+  /*< private >*/
+  gpointer	dummy1;
+  guint8	dummy2;
+  gpointer	dummy3;
+  gpointer	dummy4;
+  gpointer	dummy5;
+  gpointer	dummy6;
+};
```
pwithnall, line 63:
```suggestion
typedef struct
{
  /*< private >*/
  gpointer	dummy1;
  guint8	dummy2;
  gpointer	dummy3;
  gpointer	dummy4;
  gpointer	dummy5;
  gpointer	dummy6;
} XbNodeIter;
```

src/xb-node.h
```diff
@@ -49,6 +49,19 @@ typedef enum {
 	XB_NODE_EXPORT_FLAG_LAST
 } XbNodeExportFlags;
 
+typedef struct _XbNodeIter XbNodeIter;
+
+struct _XbNodeIter
+{
+  /*< private >*/
+  gpointer	dummy1;
+  guint8	dummy2;
```
pwithnall, line 58: Alignment

pwithnall:
> I called the iterator XbNodeIter, because I think it may be legit to repurpose the same iterator for iterating over the node's children as well (recycling the iter struct, but using two new functions to do the iteration).

I’m not sure that’s a good idea. I think you should have two separate structs, one for attributes and one for nodes. They operate on different types, so there would be no common functions between them. By reusing the same struct, you lose the ability to detect certain classes of programming error (mixing an iterator over attributes and nodes) at compile time, and checking for them at runtime would be hard/expensive. I don’t see any advantages to reusing the struct apart from saving two struct definitions (an iter and a `real` version of it).

src/xb-node.c
ximion: Weird, it looks aligned in my editor... I added an extra space, maybe that helps (it at least aligns the other entries with the name after the asterisk)

pwithnall: I think the other lines use tabs, not spaces.

src/xb-node.h
ximion: Weird, this is also aligned with tabs in my editor... It just displays weirdly on Github.

hughsie:
FWIW, I'm not worried about alignment. I'm ----><---- close to just reformatting the tree with clang-format nad never worrying about it again.

src/xb-node.c
hughsie: Ohh? Looks like I need to fix up all my projects then! Is there any doc I can reference for the commit message?

ximion:
@pwithnall @hughsie I just corrected Philip's suggestion to terminate my comment blocks with `*/` instead of `**/`, however the reason why I used the latter form in the first place was simply because all other doc comments in libxmlb seem to use that form. So that's a consistency thing, and maybe useful to change globally if it's an issue :-)

ximion:
> FWIW, I'm not worried about alignment. I'm ----><---- close to just reformatting the tree with clang-format nad never worrying about it again.

I'd actually like to switch AppStream to spaces instead of tabs (tabs make for awkward alignment sometimes, and mixed tabs-and-spaces is often strange too!) but the resulting megacommit would make Git blame annoying on a project with a long history. That's the single biggest reason why I didn't do any formatting changes.

src/xb-node.c
ximion: Me too, I also had no idea. But the double-star also make the documentation block look a bit more special than a regular, non-doc comment...

src/xb-node.h
pwithnall: :shrug:

src/xb-node.c
pwithnall: > Ohh? Looks like I need to fix up all my projects then! Is there any doc I can reference for the commit message?

I don’t think it’s worth fixing old instances of, but is just about worth ensuring new instances are right. I don’t have a reference which mentions the change, but [the examples in the gtk-doc manual](https://developer-old.gnome.org/gtk-doc-manual/stable/documenting_syntax.html.en) all use `*/`.

It’s really not important

pwithnall (commented):
LGTM now :D

ximion:
> LGTM now :D

There's one thing which I am currently wondering about: The CI failed when testing the map file:
```
1/2 libxmlb-exported-api        FAIL            0.00s   exit status 1
>>> MALLOC_PERTURB_=207 /usr/bin/diff -urNp /tmp/libxmlb/src/libxmlb.map src/libxmlb.map
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
--- /tmp/libxmlb/src/libxmlb.map        2021-11-25 14:47:52.592989287 +0100
+++ src/libxmlb.map     2021-11-25 16:22:11.170507979 +0100
@@ -249,10 +249,3 @@ LIBXMLB_0.3.1 {
     xb_builder_node_tokenize_text;
   local: *;
 } LIBXMLB_0.3.0;
-
-LIBXMLB_0.3.4 {
-  global:
-    xb_node_attr_iter_init;
-    xb_node_attr_iter_next;
-  local: *;
-} LIBXMLB_0.3.1;
```
But the Gir file has the new interfaces (they are properly introspectable), and I changed nothing which should have impacted this. Currently looking into it, but if you have an immediate idea I can maybe skip a lot of fault finding :-)

pwithnall:
You should start by looking at why `generate-version-script.py` is not generating the right map file from the GIR file, given that the GIR file is apparently correct.

hughsie:
@ximion can you rebase against main please, then LGTM. Thanks.

ximion:
Done and all CI that isn't FreeBSD seems happy :-)

PR #101, opened 2021-11-25
Include records in the symbol export list

hughsie:
This syncs with the fixed copy from fwupd.

### DIFF
--- a/src/generate-version-script.py
+++ b/src/generate-version-script.py
@@ -6,15 +6,17 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
 import sys
+import argparse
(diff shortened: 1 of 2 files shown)


hughsie:
@pwithnall sanity check on the added symbols in the .map file please.

ximion:
The quoting style change make this a bit hard to read, but the change makes sense to me! (did you format the code with Black? Looks a bit like that style!)

hughsie:
Ya, black FTW.

pwithnall (approved):
Yeah, the quoting style change should probably be a separate commit so it can be ignored during review.

The map changes look OK to me ottomh

PR #102, opened 2021-11-26
ci: Switch to new freebsd-vm action to fix build

ximion:
Hi!
I am no expert on this at all, but I like green CI lights ;-)
Simply switching to the latest bugfix release of the FreeBSD GitHub action appears to resolve the FreeBSD CI setup failure.
(the CI then is *really* slow though, I am not sure if it is expected to be that slow).


hughsie:
I see `sudo: vboxmanage: command not found` :(

ximion:
> I see `sudo: vboxmanage: command not found` :(

Yeah, weird! This didn't fail in my repo, and I just rescheduled the one from this PR and now it works.
So rather the title of this OR should be "increase probability of a working build" rather then "fix" :-/

ximion:
> I see `sudo: vboxmanage: command not found` :(

Yeah, weird! This didn't fail in my repo, and I just rescheduled the one from this PR and now it works.
So rather the title of this OR should be "increase probability of a working build" rather than "fix" :-/

ximion:
So, according to https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md and https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md, the latest version 11 of the macOS images has removed VirtualBox. So let's see if pinning the host image version actually helps to build this reliably.

PR #103, opened 2021-11-26
Add an iterator for XbNode children

ximion:
Hi!
This is the same style of iterator for XbNode children that ican now also be used for node attributes.
It is more efficient than using the `get_children()` method, and may be more obvious to use than the combination of `get_next()` and `get_child()` (while probably not being faster than those).

### DIFF
--- a/src/xb-node.c
+++ b/src/xb-node.c
@@ -50,6 +50,30 @@ G_STATIC_ASSERT (sizeof (XbNodeAttrIter) == sizeof (RealAttrIter));
 G_STATIC_ASSERT (G_ALIGNOF (XbNodeAttrIter) >= G_ALIGNOF (RealAttrIter));
 
 /**
+ * XbNodeChildIter:
+ *
+ * A #XbNodeChildIter structure represents an iterator that can be used
+ * to iterate over the children of a #XbNode. #XbNodeChildIter
+ * structures are typically allocated on the stack and then initialized
+ * with xb_node_child_iter_init().
+ *
+ * Since: 0.3.4
+ */
+
+typedef struct
+{
+  XbNode 	*node;
+  XbSiloNode	*position;
+  gboolean	 first_iter;
+  gpointer	 dummy4;
+  gpointer	 dummy5;
+  gpointer	 dummy6;
+} RealChildIter;
+
+G_STATIC_ASSERT (sizeof (XbNodeChildIter) == sizeof (RealChildIter));
+G_STATIC_ASSERT (G_ALIGNOF (XbNodeChildIter) >= G_ALIGNOF (RealChildIter));
+
+/**
  * xb_node_get_data:
  * @self: a #XbNode
  * @key: a string key, e.g. `fwupd::RemoteId`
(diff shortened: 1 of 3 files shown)


src/xb-node.c
```diff
@@ -258,6 +284,77 @@ xb_node_get_children (XbNode *self)
...
+ *
+ * Since: 0.3.4
+ */
+gboolean
+xb_node_child_iter_next (XbNodeChildIter *iter, XbNode **child)
+{
+	XbNodePrivate *priv;
+	RealChildIter *ri = (RealChildIter *) iter;
+
+	g_return_val_if_fail (iter != NULL, FALSE);
```
hughsie, line 341: g_return_val_if_fail (child != NULL, FALSE);

too?

ximion: Probably - calling this without a child parameter would be nonsensical.

hughsie: and would explode :)

hughsie:
LGTM, @pwithnall do you want to check too?

src/xb-node.c
```diff
@@ -49,6 +49,32 @@ typedef struct
...
 
+/**
+ * XbNodeChildIter:
+ *
+ * A #XbNodeChildIter structure represents an iterator that can be used
+ * to iterate over the children of a #XbNode. #XbNodeChildIter
+ * structures are typically allocated on the stack and then initialized
+ * with xb_node_child_iter_init().
+ *
+ * The iteration order of a #XbNodeChildIter is not defined.
```
pwithnall, line 60: Is it correct that the iteration order is not defined? Unlike attributes, the child nodes in an XML element tree are ordered, and order does matter in some XML schemas (like HTML).

src/xb-node.c
```diff
@@ -258,6 +284,76 @@ xb_node_get_children (XbNode *self)
...
+ * The retrieved #XbNode child needs to be dereferenced with g_object_unref().
+ * Example:
+ * |[<!-- language="C" -->
+ * XbNodeChildIter iter;
+ * g_autoptr(XbNode) child = NULL;
+ *
+ * xb_node_child_iter_init (&iter, node);
+ * while (xb_node_child_iter_next (&iter, &child)) {
+ *     // do something with the node child
+ *     g_clear_pointer (&child, g_object_unref);
```
pwithnall, line 327: You might want to also consider adding something similar to `g_variant_iter_loop()`. It’s a version of `g_variant_iter_next()` which handles the memory management of `child` for you, which saves making users remember to add this `g_clear_pointer()` call.

It’s not usable from languages other than C, though, which is why `g_variant_iter_next()` still exists. You might want to consider a similar arrangement here.

src/xb-node.c
ximion: True - the XbSilo keeps the order, and we iterate in-order, so that comment isn't actually correct.

src/xb-node.c
ximion: Interesting! I think I would need to carry an extra bool around with that then, to know the first invocation of that loop function...

ximion:
@pwithnall Please have a look at the new `xb_node_child_iter_loop` function. I don't think it's as useful as `g_variant_iter_loop`, but it will have its convenience usecases. That function could in theory call `xb_node_child_iter_next` internally, but that would result in a bit  of excess validation of known-good values (but you may prefer that).

src/xb-node.c
```diff
@@ -258,6 +282,134 @@ xb_node_get_children (XbNode *self)
...
+	*child = xb_silo_create_node (priv->silo, ri->position, FALSE);
+	ri->position = xb_silo_get_next_node (priv->silo, ri->position);
+
+	return TRUE;
+}
+
+/**
+ * xb_node_child_iter_loop: (skip)
+ * @iter: an initialized #XbNodeAttrIter
+ * @child: (out) (optional) (not nullable): Destination of the returned child
```
pwithnall, line 359:
```suggestion
 * @child: (out) (optional) (nullable): Destination of the returned child
```

It is nullable on the last iteration

pwithnall (commented):
LGTM apart from this one comment

PR #104, opened 2021-11-28
Add xb_builder_add_default_locales()

### LINKED ISSUE #41: XbBuilder could work without explicit xb_builder_add_locale() calls
Currently one must call `xb_builder_add_locale()` on an XbBuilder object to add system locales before using `xb_builder_compile()`, at least if one is using `XB_BUILDER_COMPILE_FLAG_SINGLE_LANG`. @pwithnall proposed an alternative way for it to work [here](https://gitlab.gnome.org/GNOME/gnome-software/merge_requests/405#note_690698):

> Seems a bit odd that most instances of XbBuilder have to have this loop called on them, using the results of g_get_language_names(), before they will work properly.
> Looking at the implementation, it basically checks any node which has an xml:lang attribute set. If the attribute value is not in the set of locales, that node is ignored.
> Seems like it would be faster and more convenient to have an XbBuilder:use-system-locales property, which tells XbBuilder to check for set inclusion against g_get_language_names() internally. Then you don’t need to set it up, or copy a list of strings.
> In any case, this code looks correct and what I’m suggesting is just a tentative follow-up MR at best.

hughsie:
Fixes https://github.com/hughsie/libxmlb/issues/41

### DIFF
--- a/src/libxmlb.map
+++ b/src/libxmlb.map
@@ -263,6 +263,7 @@ LIBXMLB_0.3.1 {
 
 LIBXMLB_0.3.4 {
   global:
+    xb_builder_add_default_locales;
     xb_node_attr_iter_init;
     xb_node_attr_iter_next;
   local: *;
--- a/src/xb-builder.c
+++ b/src/xb-builder.c
@@ -684,6 +684,23 @@ xb_builder_add_locale (XbBuilder *self, const gchar *locale)
 	xb_builder_append_guid (self, locale);
 }
 
+/**
+ * xb_builder_add_default_locales:
+ * @self: a #XbSilo
+ *
+ * Adds the system locales to the builder.
+ *
+ * Since: 0.3.4
+ **/
+void
+xb_builder_add_default_locales (XbBuilder *self)
+{
+	const gchar *const *locales = g_get_language_names ();
+	g_return_if_fail (XB_IS_BUILDER (self));
+	for (guint i = 0; locales[i] != NULL; i++)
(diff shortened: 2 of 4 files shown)


pwithnall:
What’s implemented here is not what was discussed in #41. Howcome?

hughsie:
I don't think we can do it that way; we need the current system locales to be part of the silo invalidation GUID.

pwithnall:
I don’t follow how that precludes adding a `use-system-locales` property and allowing it to be set to `TRUE` when constructing an `XbBuilder`.

In the `xb-tool.c` changes in this MR, that would mean:
 * `xb_builder_new()` → `xb_builder_new_with_system_locales()` or something equivalent
 * `xb_builder_constructed()` would call `xb_builder_add_default_locales()` if `use-system-locales` was set to `TRUE`
 * Explicit call to `xb_builder_add_default_locales()` in `xb-tool.c` is no longer needed

hughsie:
> xb_builder_new() → xb_builder_new_with_system_locales() or something equivalent

That seems like we're painting ourselves into a corner. What about  xb_builder_new_full() with some flags like XB_BUILDER_FLAG_USE_SYSTEM_LOCALES?

pwithnall:
Sure, the suggestion of `xb_builder_new_with_system_locales()` was just to illustrate my point

hughsie:
I made it work in the common case in 47acec2212e9c8ea3fab29f906a880b15e7dd011

PR #105, opened 2021-11-29
Allow removing XbBuilderNode text

hughsie:
This also stops the builder from exploding if the text has been removed
internally.


ximion:
I like the "file bugs for you and also immediately fix them" service! 5/5, would recommend :-P
Thank you a lot, this resolves the issue and the patch looks good to me as well!

PR #106, opened 2021-11-30
Put tail after the node when exporting XbBuilderNode

ximion:
Hi!
I thought I was going crazy yesterday (and that's still a possibility...) but now I am pretty sure that XbBuilderNode was generating wrong XML the whole time with the tail being placed inside the node rather than after.
This patch should resolve that issue, please check if that's really the issue or rather me misunderstanding something ;-)

Cheers,
    Matthias

### DIFF
--- a/src/xb-builder-node.c
+++ b/src/xb-builder-node.c
@@ -1149,12 +1149,6 @@ xb_builder_node_export_helper (XbBuilderNode *self,
 		helper->level--;
 	}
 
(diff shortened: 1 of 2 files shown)


hughsie:
@ximion ohh random. Can you add some self tests to catch things like this in the future please? e.g. just create a few XbBuilderNode's with tail text and then export them and check for the right thing? Thanks.

hughsie (approved):

PR #108, opened 2021-12-02
Wip/hughsie/quebracabezas

hughsie:


ximion:
It would probably be nice to produce a warning if the API user exceeded the token limit, rather than swallowing them and making people wonder where the tokens went... But otherwise this fixes the issue!

Issue #110, opened 2021-12-18
Non-reproducible GUID in compressed XML files


rclobus:
Hello Richard Hughes,

While working on the “reproducible builds” effort [1], I have noticed that the content of the GUID field of the converted, compressed XML files cannot be recreated in any way.

The function `xb_builder_import_node` (which is called by `as_cache_components_to_internal_xb` in appstream) uses the address of a pointer for the creation of the GUID field [2].
See the currents tests [3] for Debian, and the investigations I previously did [4].

Steps to reproduce:
* Run `appstreamcli refresh-cache`
* Notice that the content of `/var/cache/app-info/cache/C-os-catalog.xb` changes only in the GUID field (using `xb-tool`)
* If the file `C-local-metainfo.xb` is deleted before running `appstreamcli refresh-cache`, it will also differ only in the GUID field.

The tool `xb-tool` does not call `xb_builder_import_node`, so I cannot give an example that only uses libxmlb.

My goal:
* Create a live image that can be reproduced bit-by-bit.

Proposal:
* For a drop-in-replacement of the GUID field (128 bits) I see several easy alternatives:
* 1. Zero-out the field and not use it (which is probably not desired)
* 2. Calculate the MD5 checksum (which is also 128 bits) of the compressed bitstream before writing to disk
* 3. Generate a truly random UUID, with `uuid_generate_random` from libuuid (which I can de-randomize using my LD_PRELOAD-hack)
* Alternatively (but that requires more bits), something similar to GNU_BUILD_ID can be used
* If you are interested, I can provide you with a patch.

With kind regards,
Roland Clobus

 [1]: https://wiki.debian.org/ReproducibleBuilds (Reproducible Builds)
 [2]: https://sources.debian.org/src/libxmlb/0.3.6-2/src/xb-builder.c/#L649 (Using %p)
 [3]: https://jenkins.debian.net/view/live/ (Debian Jenkins)
 [4]: https://wiki.debian.org/ReproducibleInstalls/LiveImages (Section `Non-reproducible issue in Debian sid`)

ximion:
If libxmlb allowed me to reset the guid, appstream could also compute its own guid (possibly based on the files that went into the xmlb creation and their ctimes). At the moment, AS only adds its version number to the guid.
Alternatively, option 3 (completely random ID) makes sense to me to add if libxmlb `xb_builder_import_node` was called, rather than taking a pointer address.

hughsie:
[...]

PR #111, opened 2021-12-19
Ensure reproducible results when importing a node

rclobus:
Instead of using the address of a pointer, use a deterministic hash
value based on the content of the node.

Fixes https://github.com/hughsie/libxmlb/issues/110

### DIFF
--- a/src/xb-builder.c
+++ b/src/xb-builder.c
@@ -646,11 +646,12 @@ void
 xb_builder_import_node (XbBuilder *self, XbBuilderNode *bn)
 {
 	XbBuilderPrivate *priv = GET_PRIVATE (self);
-	g_autofree gchar *guid = g_strdup_printf ("bn@%p", bn);
+	g_autofree gchar *data = xb_builder_node_export(bn, XB_NODE_EXPORT_FLAG_NONE, NULL);
+	g_autofree gchar *hash = g_strdup_printf ("bn@%x", g_str_hash(data));
 	g_return_if_fail (XB_IS_BUILDER (self));
 	g_return_if_fail (XB_IS_BUILDER_NODE (bn));
 	g_ptr_array_add (priv->nodes, g_object_ref (bn));
-	xb_builder_append_guid (self, guid);
+	xb_builder_append_guid (self, hash);
 }
 
 /**


ximion:
This looks like it will be ultra-slow. The builder node export process to string is really not fast, and `import_node` may be called quite often.
I can have a look at the impact in AppStream.

hughsie:
> import_node may be called quite often

I was also worried about performance. Maybe we do both? i.e. add a xb_builder_import_node_full() which does the auto-export thing if you leave the guid string as NULL and then you can pass `""` if you don't want anything added, otherwise something sane.

rclobus:
As far as I can see, there is only one invocation in `appstreamcli`: see 
https://codesearch.debian.net/search?q=xb_builder_import_node

Refreshing the appdata database on my computer (using a GNOME desktop on Debian) takes about 0.2 second longer compared to the `%p` variant (i.e. about 10 percent slower).

rclobus:
The `export` function was the already pre-existing function that I found. 

I think that for a proper hash value, *all* node data needs to be evaluated.
 If such data can be accessed easier (i.e. without converting to proper XML) then that would be a preferred solution.

ximion:
> I was also worried about performance. Maybe we do both? i.e. add a xb_builder_import_node_full() which does the auto-export thing if you leave the guid string as NULL and then you can pass `""` if you don't want anything added, otherwise something sane.

That sounds like a good idea to me!

With this change as-is, AppStream's cache generation is ~16% slower on my machine, which is not *terrible* but definitely notable. I dislike it especially because of all the previous work done to *not* export anything as XML text anymore (as that's slow), just to add a hidden export function back ^^
That one import invocation is also in a function that is called multiple times (but it's definitely not as bad as I thought).

I think @hughsie's suggestion is a good compromise (would annoyingly mean that all current versions of AppStream would run slower until recompiled against the more recent libxmlb though).

src/xb-builder.c
```diff
@@ -646,11 +646,12 @@ void
 xb_builder_import_node (XbBuilder *self, XbBuilderNode *bn)
 {
 	XbBuilderPrivate *priv = GET_PRIVATE (self);
-	g_autofree gchar *guid = g_strdup_printf ("bn@%p", bn);
+	g_autofree gchar *data = xb_builder_node_export(bn, XB_NODE_EXPORT_FLAG_NONE, NULL);
```
ximion, line 649: nit: style
```suggestion
	g_autofree gchar *data = xb_builder_node_export (bn, XB_NODE_EXPORT_FLAG_NONE, NULL);
```

src/xb-builder.c
```diff
@@ -646,11 +646,12 @@ void
...
+	g_autofree gchar *hash = g_strdup_printf ("bn@%x", g_str_hash(data));
```
ximion, line 650: nit: style
```suggestion
	g_autofree gchar *hash = g_strdup_printf ("bn@%x", g_str_hash (data));
```

ximion, line 650: I wonder if we really still need that `bn@` prefix here... It doesn't really matter though.

hughsie:
The "do less" aka KISS version of this is here: https://github.com/hughsie/libxmlb/pull/112

rclobus:
https://github.com/hughsie/libxmlb/pull/112 implements this PR more elegantly. However, the 2 callers of `xb_builder_import_node` might need some additional code

PR #112, opened 2021-12-20
Ensure reproducible results when importing a node

hughsie:
Do not use the address of a pointer as this breaks reproducible builds.

Fixes https://github.com/hughsie/libxmlb/issues/110

Based on a patch by Roland Clobus <rclobus@rclobus.nl>, many thanks.

### DIFF
--- a/src/xb-builder.c
+++ b/src/xb-builder.c
@@ -640,17 +640,25 @@ xb_builder_generate_guid (XbBuilder *self)
  *
  * Adds a node tree to the builder.
  *
+ * If you are manually adding dynamic data sourced from a non-static source then you MUST use
+ * xb_builder_append_guid() with the appropriate GUID value, e.g. the file name and mtime.
+ *
+ * If no appropriate value is available, the caller can use something like:
+ *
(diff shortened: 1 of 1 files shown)


hughsie:
@rclobus

ximion:
This would work for me - and would also actually work with the existing code in AppStream :-)
Not sure though if there is any other code out there that would have a problem with it (but I doubt that).

ximion (approved):

rclobus:
Now that a call to `xb_builder_append_guid` is a MUST to ensure a proper unique identifier, could you take a look at the 2 callers of `xb_builder_import_node` that I found in the Debian source code repository at https://codesearch.debian.net/search?q=xb_builder_import_node

@ximion appstream 0.15.0
https://github.com/ximion/appstream/blob/master/src/as-cache.c
@hughsie  fwupd 1.5.7
https://github.com/fwupd/fwupd/blob/main/src/fu-engine.c

hughsie:
> fwupd 1.5.7

Yes, I checked this before merging. the tl;dr: is that fwupd does `xb_builder_compile()` unconditionally rather than `xb_builder_ensure()` and so the GUID isn't so important.

ximion:
Same for AppStream, which has its own logic in front to rebuild the cache only if needed, so it isn't dependent on the GUID either.

PR #113, opened 2021-12-20
trivial: Reformat with clang-format

hughsie:
It turns out life is too short to worry about whitespace.


pwithnall (commented):
Sure. Are you planning on adding something as a pre-merge hook or CI job which forcibly reformats or checks code against the new `.clang-format` file?

You probably also want to [add a `.git-blame-ignore-revs` file](https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame) in a followup commit so that this mass-reformat can easily be ignored in `git blame` output in future.

ximion:
Of course this goes into bikeshed-territory, but: Can the clang-format configuration be altered to mimick the existing coding style as much as possible? That would reduce the delta of this patch, and may look a bit prettier (I wonder in particular whether there is an option to keep the space before the brackets, as that contributes to a lot of churn, and whether it is possible to auto-align struct members rather than bashing them together. The header files also look a bit more "cramped" and harder to read now with the return-type-linebreak-function style. Not sure if that can be solved only for header files though, and the formatter also won't know what to group together anyway).
Also, I didn't know you could ignore stuff in git blame like this! That's one of the biggest reasons for me not doing any large-scale  reformatting in projects ;-)

hughsie:
> is possible to auto-align struct members rather than bashing them together

That's not possible with clang-format yet.

> I didn't know you could ignore stuff in git blame like this

Ya, I'll add that and the precommit stuff in a followup commit.

pwithnall:
New version LGTM, although I haven’t read through the Python scripts in `contrib/`

PR #115, opened 2022-01-30
build: Require Python explicitly

jtojnar:
When the first argument to python.find_installation is omitted, Meson will return the Python instance used to run it. But that can be a completely different installation than the one on PATH.

This change was initially introduced in https://github.com/hughsie/libxmlb/commit/ca94c2ebfed0c9c370279fdba5ad4d03456be5b6 but reverted in https://github.com/hughsie/libxmlb/commit/9d520fff9aae1028ada992a3252de52c79d940b2 due to concern about FreeBSD 12.2. But FreeBSD 12.2 installs python3 on PATH nowadays and even if it did not, the inability to find Python is a [bug to fix in Meson](https://github.com/mesonbuild/meson/issues/4608).

cc @Asiderr (https://github.com/hughsie/libxmlb/pull/92)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -151,7 +151,7 @@ configure_file(
 )
 
 python = import('python')
-python3 = python.find_installation()
+python3 = python.find_installation('python3')
 
 subdir('data')
 subdir('src')


eli-schwartz:
This is wrong for two reasons:
- the correct solution is "why does this not use find_program()" #116
- >  Meson will return the Python instance used to run it. But that can be a completely different installation than the one on PATH.
  Why do you care which python it finds, if it is just being used as a script interpreter then it doesn't (should not) matter.

jtojnar:
> * > Meson will return the Python instance used to run it. But that can be a completely different installation than the one on PATH.
>       
>       
>       Why do you care which python it finds, if it is just being used as a script interpreter then it doesn't (should not) matter.

Different pythons can have different `PYTHONPATH`. Particularly, the Meson one does not have setuptools with Nix.

eli-schwartz:
I did say "should". And I have put my money where my mouth is and removed the setuptools dependency. :D

jtojnar:
In the interest of resolving the underlying question, what do you think about using `find_installation` in cases where it is used to check for a library that definitely is not available by default and it is not easy to replace it? For example, here `pygobject3` is used for running tests: https://gitlab.gnome.org/GNOME/tracker/-/merge_requests/410/diffs

eli-schwartz:
I would say that that is fine.

It's not that using the find_installation method is inherently bad or evil... I just believe it is typically used when it isn't needed, and that has side effects:
- it is slower to import, `import('foo')` has to read another file from disk and load it into the interpreter. This isn't much, but wasting a small amount of effort is still waste
- it is slower to find, because it will fork+exec a process to run some json info dumper program written in python. One fork+exec isn't that costly, but it is a lot more costly than the original module import was. And `find_program()` doesn't fork, unless you specify a version number
- find_installation can actually log warnings if it thinks that installing built modules cannot find a site-packages directory on your system PATH, that produce errors with `--fatal-meson-warnings`. Why go through that codepath if you don't get anything out of it?

All this is minor stuff that is overwhelmed by actual real needs like "if we don't check that this module exists, there are real life systems where we successfully configure, but fail to compile". That will always be the priority.

jtojnar:
Thanks for explaining.

PR #116, opened 2022-01-30
meson: use the correct lookup method for the python3 script interpreter

eli-schwartz:
The python module's find_installation method is incorrect -- it does a variety of things geared towards building and installing python modules into site-packages. This is slow and inefficient and completely pointless -- all that is needed here is to run scripts.

find_program() is guaranteed to find the python3 interpreter too, and is the correct way to find *any* other program, so use it for python as well.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,3 +4,6 @@ root = true
 [*]
 indent_style = tab
 indent_size = 8
+
+[*.py]
+indent_style = space
--- a/meson.build
+++ b/meson.build
@@ -150,8 +150,7 @@ configure_file(
   configuration : conf
 )
 
-python = import('python')
-python3 = python.find_installation()
+python3 = find_program('python3')
 
 subdir('data')
 subdir('src')
--- a/src/generate-version-script.py
+++ b/src/generate-version-script.py
@@ -9,11 +9,12 @@ import sys
 import argparse
 import xml.etree.ElementTree as ET
 
-from pkg_resources import parse_version
-
 XMLNS = "{http://www.gtk.org/introspection/core/1.0}"
 XMLNS_C = "{http://www.gtk.org/introspection/c/1.0}"
 
+
+def parse_version(ver):
+    return tuple(map(int, ver.split('.')))
 
 def usage(return_code):
     """print usage and exit with the supplied return code"""


jtojnar:
We might want to check for `xml.etree.ElementTree` like we do in libjcat:

https://github.com/hughsie/libjcat/blob/f802f078ec0349e4056bf764ee4da16fe9973978/libjcat/meson.build#L156-L162

eli-schwartz:
I find it unlikely that you need to check xml.etree.ElementTree?

It's a pure python module shipped in the standard library, any python installation without it is simply invalid, and moreover, the distributor of the binary installation would have had to go to unnatural lengths to cripple it in that fashion. And the expat parser is bundled as vendorized source code in cpython so there is no excuse for it to not be available.

Note that `xml.etree.ElementTree` uses a C accelerated `_elementtree` module for Element, SubElement, ParseError, TreeBuilder, XMLParser... this actually has explicit support in the CPython codebase for not supporting `_elementtree`, but, in that case it will just use a reference pure python implementation. Under no circumstances is it expected to be entirely missing. It's not "optional" to have this module.

Given neither this PR nor your PR actually attempt to check for the presence of the module, I see no problem with using `find_program()` here...

jtojnar:
Interesting, did not know that. But we would probably still want to check for `pkg_resources`, which is part of `setuptools` package?

eli-schwartz:
> But we would probably still want to check for `pkg_resources`, which is part of `setuptools` package?

Meson explicitly does not depend on third-party modules, which pkg_resources is, and it cannot be assumed that setuptools is installed where Meson is installed. When building on Linux, the python used for Meson as `mesonlib.python_command` and the python used as `shutil.which('python3')` are invariably the same (unless the user installs their own python); on BSD, I guess it usually is?

Setuptools is not actually supposed to be a runtime module, it's in the name: ***setup***tools. It isn't a requirement that it be installed just because packages use it to build. Moreover, `parse_version` is an aliased import for the `packaging.version` module, except that `pkg_resources` is, by design, incredibly slow (it processes every. single. package. in your entire python installation. Upfront. Then throws out all that information because you don't actually do anything with the conveniently autoloaded WorkingSet.)

So adding a build dependency on setuptools feels kind of silly. Regardless of whether this meson.build checks for it, on many systems you will need to add an extra build depends on python3-setuptools just to run a testsuite script that uses setuptools to sort version numbers... and for what??? An `ld` version script does not care in what order you define symbol mapping rules.

Either way, I am not hugely motivated to care. Just please do NOT use python_mod.find_installation for frivolous reasons, which is exactly what using it to just to find "any python3 executable, it doesn't matter which one" is.

jtojnar:
> > But we would probably still want to check for `pkg_resources`, which is part of `setuptools` package?
> 
> Meson explicitly does not depend on third-party modules, which pkg_resources is, and it cannot be assumed that setuptools is installed where Meson is installed.

Right, that is why we need to check it is present.

> When building on Linux, the python used for Meson as `mesonlib.python_command` and the python used as `shutil.which('python3')` are invariably the same (unless the user installs their own python); on BSD, I guess it usually is?

This is not true for Nix. Meson’s Python will not be on Path because we wrap it not to interfere with project’s own dependencies.

> Setuptools is not actually supposed to be a runtime module, it's in the name: _**setup**_tools. It isn't a requirement that it be installed just because packages use it to build.

Well, we do only use it for build time. It is just that I am not aware of any other way to check for Python module presence.

And we do want to check for the module presence beforehand because having build run and fail it because of missing dependency at the end is incredibly frustrating. (It might be not that much of a pain for small project like this one but this exact issue cost me hours as a packager so it is a pet peeve of mine.)

> Moreover, `parse_version` is an aliased import for the `packaging.version` module, except that `pkg_resources` is, by design, incredibly slow (it processes every. single. package. in your entire python installation. Upfront. Then throws out all that information because you don't actually do anything with the conveniently autoloaded WorkingSet.)

Did not know that, we typically have lean package-specific environment. But I see that how this might be a problem and we definitely can switch to `packaging.version` directly.

> Regardless of whether this meson.build checks for it, on many systems you will need to add an extra build depends on python3-setuptools just to run a testsuite script that uses setuptools to sort version numbers... and for what??? An `ld` version script does not care in what order you define symbol mapping rules.

I guess it makes it easier to compare the version script between versions? Not really sure about this, I am mostly just a packager trying to improve upstream packaging to the best of my knowledge.
 
> Either way, I am not hugely motivated to care. Just please do NOT use python_mod.find_installation for frivolous reasons, which is exactly what using it to just to find "any python3 executable, it doesn't matter which one" is.

It certainly does matter on Nix. And while `find_program` would find the same installation, we still need to check for at least `packaging.version` to fail fast in case of insufficient dependencies.

eli-schwartz:
> Well, we do only use it for build time. It is just that I am not aware of any other way to check for Python module presence.

Allow me to clarify my original statement. Setuptools will only be available at build time for Meson, not build time for libxmlb.

libxmlb can, of course, pull it in as an additional direct build dependency, but no one should mistakenly think that "setuptools is popular, therefore we can depend on it because it is probably preinstalled already".

> I guess it makes it easier to compare the version script between versions? Not really sure about this, I am mostly just a packager trying to improve upstream packaging to the best of my knowledge.

The version script is both checked into git and generated into the builddir using this python script. The latter is only ever used as a unittest to check that it generates the same file, apparently. But my point is that doesn't require any sorting at all, just a reproducible order, which can be accomplished by alphabetic string sorting.

> It certainly does matter on Nix. And while `find_program` would find the same installation, we still need to check for at least `packaging.version` to fail fast in case of insufficient dependencies.

The current version does not check, so the current version (perhaps erroneously) claims it doesn't matter. The best solution would be to make it *actually* not matter, by removing pkg_resources from the equation. It's not clear to me what the benefit of it is.

Perhaps @hughsie could clarify.

jtojnar:
> libxmlb can, of course, pull it in as an additional direct build dependency, but no one should mistakenly think that "setuptools is popular, therefore we can depend on it because it is probably preinstalled already".

Right, this is what I was saying as well. Hence I added explicit check for setuptools to libjcat.

> > It certainly does matter on Nix. And while `find_program` would find the same installation, we still need to check for at least `packaging.version` to fail fast in case of insufficient dependencies.
> 
> The current version does not check, so the current version (perhaps erroneously) claims it doesn't matter.

Right, it is IMO an omission. The build script should have been checking for it like we do in `libjcat`.

> The best solution would be to make it _actually_ not matter, by removing pkg_resources from the equation.

I am all for removing complexity when not needed, and as you point out, it might be the case here. Though only Richard can say here for sure.

In the interest of resolving the underlying question, what do you think about using `find_installation` in cases where it is used to check for a library that definitely is not available by default and it is not easy to replace it? For example, here `pygobject3` is used for running tests: https://gitlab.gnome.org/GNOME/tracker/-/merge_requests/410/diffs

hughsie:
> Perhaps @hughsie could clarify.

Ohh, I'm the least qualified to get involved here! :) Really I just need parse_version() and if that's available elsewhere then ideas welcome.

jtojnar:
The question is whether you really need parse_version, or if sorting alphabetically would be enough.

eli-schwartz:
Well, that's the thing. My question here is, why do you need parse_version? Why do you need to "get a sorted list of all the versions" -- why isn't an *unsorted* list equally suitable, or at least a reproducibly sorted list that doesn't implement *version* sorting?

In `_add_cls` you do compare version_lowest and version_tmp, and I don't thoroughly understand the nuances of that... but assuming the need for figuring out the lowest version is indeed real, here's a simple alternative:

You could do

```python
version_string = '1.2.3'
version_tuple = version_string.split('.')

if version_tuple > otherversion_tuple:
    ...
```

Tuple comparison of a rigidly defined version schema using only dot-separated integers and no alpha/beta/rc postfixes is trivial to implement. The `Version` class which setuptools/packaging.version uses, is a heavyweight that handles many different rules as well as having a fallback class `LegacyVersion`, the two together are supposed to handle "literally any bizarre format out there" and smooth over the differences.

While you're at it, you could use that in the version script generator too, I suppose.

eli-schwartz:
I pushed a change to not use setuptools. It seems to work, as in `ninja test` currently passes.

hughsie:
@superm1 I think we only have sanitized triplets, right? i.e. I think this approach works.

superm1:
> @superm1 I think we only have sanitized triplets, right? i.e. I think this approach works.

Yeah

jtojnar (approved):
Sounds reasonable to me, provided that all handled versions are straightforward like that.

hughsie:
Merged with some re-ordering. Thanks!

Issue #117, opened 2022-02-03
Hidden files also monitored when using XB_BUILDER_SOURCE_FLAG_WATCH_DIRECTORY


pabloyoyoista:
While debugging GNOME Software and doing upgrades of packages in alpine linux, I get (lots of) repeated debug messages of invalidation of hidden files.  Similar to those below:
```
XbSilo /usr/share/applications/.mimeinfo.cache.DLPXG1 changed, invalidating
XbSilo /usr/share/metainfo/.apk.$long_hash changed, invalidating
```
Looks like alpine's package manager, apk, is creating temporary files during upgrades. Looking at the [MR that introduced](https://github.com/hughsie/libxmlb/pull/55) the flag and the [corresponding one](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/469) in Software, it looks like that was never the intention.

The small snippet below would probably fix it, but I am not sure if this is something desirable, or if the current behavior is the intended.  So I thought I would open an issue before creating a MR.

```
diff --git i/src/xb-silo.c w/src/xb-silo.c
index 288f6b8..6fae596 100644
--- i/src/xb-silo.c
+++ w/src/xb-silo.c
@@ -878,6 +878,8 @@ xb_silo_watch_file_cb(GFileMonitor *monitor,
        g_autofree gchar *basename = g_file_get_basename(file);
        if (g_str_has_prefix(basename, ".goutputstream"))
                return;
+       if (g_str_has_prefix(basename, "."))
+               return;
        g_debug("%s changed, invalidating", fn);
        xb_silo_invalidate(silo);
 }
```

pabloyoyoista:
> Is it creating them in /usr/share/applications/ and /usr/share/metainfo?

Yes, that looks like it is the case. It is probably harmless, but if it is totally unnecessary, isn't it reasonable to avoid the unnecessary work?

PR #118, opened 2022-02-07
Ignore all hidden files when using _WATCH_DIRECTORY

hughsie:
We already ignore the GFile-default file prefix, but it's much clearer
to just ignore all hidden files.

Fixes https://github.com/hughsie/libxmlb/issues/117


pwithnall (approved):
LGTM

Issue #120, opened 2022-03-30
XbSilo-CRITICAL **: strtab+offset is outside the data range for 12


mcrha:
I do not know what I've missing here, but I surely do something miss here. When I:
1. `mkdir -p  /tmp/share/app-info/xmls`
2. `cp /usr/share/app-info/xmls/fedora-popular.xml /tmp/share/app-info/xmls`
3. compile the attached [xml-parser.c.txt](https://github.com/hughsie/libxmlb/files/8382244/xml-parser.c.txt)  (the first line contains a command to compile & run it) and run it, then I get:

```
process:66515): XbSilo-CRITICAL **: 18:03:12.609: strtab+offset is outside the data range for 12

** (process:66515): WARNING **: 18:03:12.609: Failed to compile data: strtab_ntags incorrect
```

I do not know why it does that, but when I comment out the `"RemoveNonMergeComponents"` fixup (beginning at line 142), then it is happy.

This is with the current sources, at commit 3f25c84f6f11df5f426a872c5616976f04a16870.

When I keep the ` /tmp/share/app-info/xmls` empty, then I get a crash:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7fa952a in xb_silo_get_node_element (self=0x4178e0, n=0x0) at ../src/xb-silo.c:438
438		return xb_silo_from_strtab(self, n->element_name);

#0  0x00007ffff7fa952a in xb_silo_get_node_element (self=0x4178e0, n=0x0) at ../src/xb-silo.c:438
#1  0x00007ffff7fa525c in xb_node_get_element (self=0x419020) at ../src/xb-node.c:488
#2  0x0000000000402efa in gs_appstream_traverse_silo_for_index (node=0x419020, index=0x409b00, depth=0) at xml-parse.c:190
#3  0x00000000004030f1 in gs_appstream_create_silo_index (silo=0x4178e0) at xml-parse.c:220
#4  0x0000000000403232 in main (argc=1, argv=0x7fffffffda98) at xml-parse.c:248
```

An interesting observation, when I change the path, like to `/tmp/xxx`, then the runtime warning has a different number.

For the code itself: I want to read all components, which have the `merge` attribute and discard (ignore) everything else. I want also the opposite, read only components without the `merge` attribute. Maybe I do something wrong, or the `XB_BUILDER_NODE_FLAG_IGNORE` flag misbehaves, or anything else.

mcrha:
[...]

PR #121, opened 2022-03-31
Fix a crash for when the root tree has no children

hughsie:
This will only happen when using xb_node_get_child(xb_silo_get_root()).

Fixes https://github.com/hughsie/libxmlb/issues/120

### DIFF
--- a/src/xb-builder.c
+++ b/src/xb-builder.c
@@ -254,6 +254,15 @@ xb_builder_compile_source(XbBuilderCompileHelper *helper,
 	if (!xb_builder_source_fixup(source, root_tmp, error))
 		return FALSE;
 
+	/* check to see if the root was ignored */
+	if (xb_builder_node_has_flag(root_tmp, XB_BUILDER_NODE_FLAG_IGNORE)) {
+		g_set_error_literal(error,
+				    G_IO_ERROR,
+				    G_IO_ERROR_INVALID_DATA,
+				    "root node cannot be ignored");
+		return FALSE;
+	}
+
 	/* a single root with no siblings was required */
 	if (helper->compile_flags & XB_BUILDER_COMPILE_FLAG_SINGLE_ROOT) {
 		if (xb_builder_node_get_children(root_tmp)->len > 1) {
--- a/src/xb-builder.c
+++ b/src/xb-builder.c
@@ -280,7 +280,8 @@ xb_builder_compile_source(XbBuilderCompileHelper *helper,
 		children = xb_builder_node_get_children(helper->current);
 		for (guint i = 0; i < children->len; i++) {
 			XbBuilderNode *bn = g_ptr_array_index(children, i);
-			xb_builder_node_add_child(bn, info);
+			if (!xb_builder_node_has_flag(bn, XB_BUILDER_NODE_FLAG_IGNORE))
+				xb_builder_node_add_child(bn, info);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


hughsie:
@mcrha can you verify please.

mcrha:
When you apply this [xb.patch.txt](https://github.com/hughsie/libxmlb/files/8390167/xb.patch.txt) on top of this merge request, then it'll reproduce the crash.

As the `xb_silo_get_root` returns a valid node, it should be able to use all the `XbNode` methods, right?

mcrha:
And this patch [xb.patch.txt](https://github.com/hughsie/libxmlb/files/8390284/xb.patch.txt) extends the previous patch to reproduce the `strtab incorrect` error - that might need a new test, I guess.

src/xb-silo.c
```diff
@@ -435,6 +435,8 @@ xb_silo_get_node_tail(XbSilo *self, XbSiloNode *n)
 const gchar *
 xb_silo_get_node_element(XbSilo *self, XbSiloNode *n)
 {
+	if (n == NULL)
+		return NULL;
```
mcrha, line 439: I believe it'll be a good idea to be proactive and to catch all the `xb_node_get_` functions in this merge request, not only those I happened to use.

mcrha:
Still not perfect:
```
Thread 36 "gnome-software" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff51ca3640 (LWP 18027)]
0x00007ffff6e14ac4 in xb_silo_node_has_flag (self=0x0, flag=XB_SILO_NODE_FLAG_IS_ELEMENT) at ../src/xb-silo-node.c:39
39		return (self->flags & flag) > 0;
#0  0x00007ffff6e14ac4 in xb_silo_node_has_flag (self=0x0, flag=XB_SILO_NODE_FLAG_IS_ELEMENT) at ../src/xb-silo-node.c:39
#1  0x00007ffff6e14a6e in xb_silo_node_get_size (self=0x0) at ../src/xb-silo-node.c:17
#2  0x00007ffff6e0feca in xb_silo_get_child_node (self=0xd41df0, n=0x0) at ../src/xb-silo.c:277
#3  0x00007ffff6e0be66 in xb_node_child_iter_init (iter=0x7fff51ca24f0, self=0x7fff580e4040) at ../src/xb-node.c:306
```

hughsie:
> Still not perfect

Aha, of course. One more commit to test please.

mcrha:
```
Thread 35 "gnome-software" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff524a4640 (LWP 63838)]
0x00007ffff6e0fe8a in xb_silo_get_next_node (self=0xd42600, n=0x0) at ../src/xb-silo.c:266
266		if (n->next == 0x0)
#0  0x00007ffff6e0fe8a in xb_silo_get_next_node (self=0xd42600, n=0x0) at ../src/xb-silo.c:266
#1  0x00007ffff6e0bc75 in xb_node_get_next (self=0x7fff580e3510) at ../src/xb-node.c:227
```

It would really be beneficial to cover all the `priv->silo, priv->sn` usages, otherwise it's just that it'll work for calls my code does, but can be broken for someone else, using different calls.

hughsie:
> It would really be beneficial to cover all the priv->silo, priv->sn usages

Okay, agreed. Done.

mcrha:
This works fine here, no crash spotted so far. Thank you.

PR #125, opened 2022-05-13
xb-builder: Watch files before loading them

pwithnall:
This eliminates a slight race where the file is loaded, then could be
modified by another process before it’s watched.

See https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1517

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>


pwithnall:
I have not tested this.

pwithnall:
> I have not tested this.

Evidently the CI thinks that was a bad idea and I should put more time into this.

PR #127, opened 2022-09-02
xb_builder: fix double free

sergio-costas:
In some cases there can be a double free of XbBuilderNode
elements. This can happen when a XbBuilderNode is marked to
be destroyed when including only the highest priority
translation, if any of its childer is also marked to be
destroyed there too. In that case, the children is destroyed
twice: one when the parent is destroyed, and another when
it is destroyed in the loop when

    flags & XB_BUILDER_COMPILE_FLAG_SINGLE_LANG

is true.

To fix this, this patch increases the refcount of each element
added to the 'nodes_to_destroy' ptr_array in the
'xb_builder_xml_lang_prio_cb' function, and decrements it after
destroying it in the loop. This guarantees that no element is
destroyed twice.

This bug was discovered because it caused a segmentation fault
in Gnome Software. Applying this patch fixes it, and removes
several errors detected by Valgrind.


src/xb-builder.c
```diff
@@ -852,6 +852,7 @@ xb_builder_compile(XbBuilder *self,
 		for (guint i = 0; i < nodes_to_destroy->len; i++) {
 			XbBuilderNode *bn = g_ptr_array_index(nodes_to_destroy, i);
 			xb_builder_node_unlink(bn);
+			g_object_unref(bn);
```
hughsie, line 855: could we define nodes_to_destroy with `g_ptr_array_new_with_free_func((GDestroyNotify)g_object_unref);` to avoid this?

sergio-costas: Of course!

sergio-costas: (just testing it with gnome software before sending the patch)

sergio-costas: Done.

sergio-costas:
How can I see the error in the "Codacy Static Code Analysis"?

hughsie:
> How can I see the error in the "Codacy Static Code Analysis"?

That's my problem; it was looking for master, and now it's main. I've fixed it.

hughsie (approved):
Many thanks for this!

PR #131, opened 2023-01-29
Use indexes when binding value

hughsie:
This means we still use the index even when using xb_value_bindings_bind_str().


src/xb-machine.c
```diff
@@ -1559,17 +1562,28 @@ xb_machine_func_eq_cb(XbMachine *self,
 					  error);
 
 	/* TEXT:TEXT */
-	if (xb_opcode_cmp_str(&op1) && xb_opcode_cmp_str(&op2))
+	if (xb_opcode_cmp_str(&op1) && xb_opcode_cmp_str(&op2)) {
+		if (priv->debug_flags & XB_MACHINE_DEBUG_FLAG_SHOW_SLOW_PATH) {
+			gchar *str1 = xb_opcode_to_string(&op1);
```
pwithnall, line 1567: Either this is a leak, or `str1` should be `const`. Same below in a few places.

src/xb-machine.h
```diff
@@ -35,6 +35,7 @@ struct _XbMachineClass {
  * @XB_MACHINE_DEBUG_FLAG_SHOW_STACK:		Show the stack addition and removal
  * @XB_MACHINE_DEBUG_FLAG_SHOW_PARSING:		Show the XPath predicate parsing
  * @XB_MACHINE_DEBUG_FLAG_SHOW_OPTIMIZER:	Show the optimizer operation
+ * @XB_MACHINE_DEBUG_FLAG_SHOW_SLOW_PATH:	Show the queries slow paths
```
pwithnall, line 38:
```suggestion
 * @XB_MACHINE_DEBUG_FLAG_SHOW_SLOW_PATH:	Show the query’s slow paths
```

Or “queries’”

src/xb-value-bindings.c
```diff
@@ -17,30 +17,23 @@
...
-	enum {
-		KIND_NONE,
-		KIND_TEXT,
-		KIND_INTEGER,
-	} kind;
-	union {
-		gchar *text;
-		guint32 integer;
-	};
+	XbOpcodeKind kind;
```
pwithnall, line 24: Using `XbOpcodeKind` here seems risky, given it has a domain of values much wider than what’s needed here. There’s now no protection in the type system from ending up with `XB_OPCODE_KIND_INTEGER` rather than `XB_OPCODE_KIND_BOUND_INTEGER`, for example.

src/xb-value-bindings.c
```diff
@@ -17,30 +17,23 @@
...
 typedef struct {
-	enum {
-		KIND_NONE,
-		KIND_TEXT,
-		KIND_INTEGER,
-	} kind;
-	union {
-		gchar *text;
-		guint32 integer;
-	};
```
pwithnall, line 31: Why get rid of the union? Now the struct is a bit bigger, which is important because it’s supposed to be used on the stack. The bigger it is, the less likely it’ll fit in L1, hence it might be slower.

pwithnall (commented):
Just a quick review, sorry, I’ve not touched this code in a while and don’t have any time at work in the next 6 months for work on libxmlb.

Feel free to ignore any of my comments since I can’t get pulled into a protracted review discussion. Sorry!

src/xb-machine.h
hughsie: It's really one query -- fixed, thanks.

src/xb-machine.c
hughsie: Good catch, thanks, fixed.

src/xb-value-bindings.c
hughsie: Agreed, I've switched it back.

src/xb-value-bindings.c
pwithnall: OK, makes sense

PR #134, opened 2023-04-26
disable zstd support added

sulincix:
This changes allow to disable zstd support.

Default status enabled.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,7 +124,7 @@ if get_option('zstd')
     if not zstd.found() 
         error('libzstd is missing, please install it to continue!')
     endif
-      conf.set('HAVE_ZSTD', 1)
+        conf.set('HAVE_ZSTD', 1)
 endif
 if giounix.found()
   conf.set('HAVE_GIO_UNIX', '1')
--- a/meson.build
+++ b/meson.build
@@ -119,6 +119,13 @@ mandir = join_paths(prefix, get_option('mandir'))
 gio = dependency('gio-2.0', version : '>= 2.45.8')
 giounix = dependency('gio-unix-2.0', version : '>= 2.45.8', required: false)
 lzma = dependency('liblzma')
+zstd = dependency('libzstd')
+if get_option('zstd')
+    if not zstd.found() 
+        error('libzstd is missing, please install it to continue!')
+    endif
+      conf.set('HAVE_ZSTD', 1)
+endif
 if giounix.found()
   conf.set('HAVE_GIO_UNIX', '1')
 endif
@@ -134,8 +141,7 @@ libxmlb_deps = [
   lzma
 ]
 if get_option('zstd')
-    libxmlb_deps += dependency('libzstd')
-    add_project_arguments('-DENABLE_ZSTD', language: 'c')
+    libxmlb_deps += zstd
 endif
 # support stemming of search tokens
 if get_option('stemmer')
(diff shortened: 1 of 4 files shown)


hughsie:
Ohh wow, you've got an environment without zstd?

sulincix:
Yes. I am not using zstd :D

src/meson.build
```diff
@@ -39,6 +39,11 @@ subdir('libxmlb')
 
 mapfile = 'libxmlb.map'
 vflag = '-Wl,--version-script,@0@/@1@'.format(meson.current_source_dir(), mapfile)
+if get_option('zstd')
+    zstd_sources = ['xb-zstd-decompressor.c']
+else
+    zstd_sources = []
```
hughsie, line 45: maybe call this extra_sources or something? I could imagine other things there too.

sulincix: Okay. I am changing now.

hughsie: also, maybe define the [] before the `if get_op` and then use += for the xb-zstd-decompressor

sulincix: https://github.com/hughsie/libxmlb/pull/134/commits/9a751e1576b8fa68d8fe3b41a8d81d0341fd6e3f fixed

sulincix: I think if we will add extra features. we can use extra_sources array. 
Also extra_sources used 2 times in file...

meson.build
```diff
@@ -119,7 +119,6 @@ mandir = join_paths(prefix, get_option('mandir'))
 gio = dependency('gio-2.0', version : '>= 2.45.8')
 giounix = dependency('gio-unix-2.0', version : '>= 2.45.8', required: false)
 lzma = dependency('liblzma')
-zstd = dependency('libzstd')
```
ximion, line 122: You could simplify the code by just using something like this:
``` meson
zstd = dependency('libzstd')
if get_option('zstd')
    add_project_arguments('-DENABLE_ZSTD', language: 'c')
    if not zstd.found() 
        error('libzstd is missing, please install it to continue!')
    endif
endif
``` 
That avoids having the extra `libxmlb_deps` variable.
By adding a `HAVE_ZSTD` define directly to the project's `configuration_data` above, you can simplify the code even more and have an even smaller Meson definition (with only `extra_sources remaining).

hughsie: I think you could use a meson `feature` too.

sulincix: https://github.com/hughsie/libxmlb/pull/134/commits/f72a109986f0d2ad6dedf4239a7489272bd24b21

meson.build
```diff
@@ -120,6 +120,12 @@ gio = dependency('gio-2.0', version : '>= 2.45.8')
 giounix = dependency('gio-unix-2.0', version : '>= 2.45.8', required: false)
 lzma = dependency('liblzma')
 zstd = dependency('libzstd')
+if get_option('zstd')
+    if not zstd.found() 
+        error('libzstd is missing, please install it to continue!')
+    endif
+      conf.set('HAVE_ZSTD', 1)
```
hughsie, line 127: i think the indenting is wonky here -- two less spaces

sulincix: https://github.com/hughsie/libxmlb/blob/main/meson.build#L144 I wrote like this

sulincix: https://github.com/hughsie/libxmlb/pull/134/commits/a074f2e283c2333a317f3fbb7fad4d1677fea4da fixed.

meson.build
```diff
@@ -120,6 +120,12 @@ gio = dependency('gio-2.0', version : '>= 2.45.8')
 giounix = dependency('gio-unix-2.0', version : '>= 2.45.8', required: false)
 lzma = dependency('liblzma')
 zstd = dependency('libzstd')
+if get_option('zstd')
+    if not zstd.found() 
+        error('libzstd is missing, please install it to continue!')
```
eli-schwartz, line 125: But zstd is mandatory by default... this is why a feature option was suggested.

PR #136, opened 2023-06-04
Update meson.build to fix compile issue

taozuhong:
The build script throw error with MSYS2 on Windows:
```
meson.build:117:20: ERROR: Unknown variable "prefix".
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -102,12 +102,12 @@ add_project_link_arguments(
   language: 'c'
 )
 
+prefix = get_option('prefix')
 if host_machine.system() == 'windows'
   bindir = get_option('bindir')
   installed_test_bindir = get_option('libexecdir')
   installed_test_datadir = get_option('datadir')
 else
-  prefix = get_option('prefix')
   datadir = join_paths(prefix, get_option('datadir'))
   bindir = join_paths(prefix, get_option('bindir'))
   libexecdir = join_paths(prefix, get_option('libexecdir'))


hughsie:
I don't think that's right; we don't want to use the prefix on win32. Two suggestions:

 * move the `mandir = join_paths(prefix, get_option('mandir'))` into each block, and use `mandir = join_paths(get_option('mandir'))` for the windows one
 * Make manpages conditional on non-Windows -- I don't think they make sense in the win32 deliverable anyway.

Thanks!

elmarco:
> I don't think that's right; we don't want to use the prefix on win32. Two suggestions:
>

why not? btw, iirc meson takes care of ignoring prefix if mandir is absolute anyway (not verified)

>     * move the `mandir = join_paths(prefix, get_option('mandir'))` into each block, and use `mandir = join_paths(get_option('mandir'))` for the windows one

probably not necessary

>     * Make manpages conditional on non-Windows -- I don't think they make sense in the win32 deliverable anyway.

maybe, maybe not.. I don't think people really care judging by the fact that I can't see a "man" packaged in msys. Yet a good chunk of packages install man pages. ¯\_(ツ)_/¯ 

in the mean time, I think this patch should be applied :)

taozuhong:
The document tool using the `gtkdoc` in the file `meson_options.txt`, and keep path spec consistently is good for developers.

So I think that apply this PR is fine.

hughsie (approved):
Okay, makes sense. Thanks.

PR #137, opened 2023-07-31
xb-silo: skip xb_query_to_string() unless error is set

chergert:
If xb_silo_query_with_root_full() is called without a GError, then there is no need to format the query for a GError that will be discarded.

In GNOME Software, for example, this could be called 70,000 times for a trivial query via GNOME Shell. Avoiding this saves about 4% of CPU on an idle system transitioning to Shell search.


chergert:
My pleasure!

Issue #143, opened 2023-08-14
Strange behavior when searching for tokens


ximion:
Hi!

I was debugging some strange behavior in AppStream which may actually be an issue with libxmlb (or at the very least I need some guidance on what the correct approach is here).

AppStream performs its own stemming, so a term like `strategy` is stemmed to `strategi` (making it no longer a fulltext search match). AppStream adds the stemmed terms to a node using `xb_builder_node_add_token ();`.
I have verified that the tokens are actually added, they show up in `xb-tool dump` output in the correct places.

Now, I first perform a query to get all component nodes in the XML, like this: `cpt_nodes = xb_silo_query (csec->silo, "components/component", 0, &tmp_error);`

A new query is constructed like this:
``` c
query = xb_query_new (csec->silo,
			queries[j].xpath,
			&error_query);
``` 
With XPath `summary[text()~=?]`

Then I iterate over all component nodes `cpt_node` with search term `term` (a single, stemmed word):
``` c
g_auto(XbQueryContext) context = XB_QUERY_CONTEXT_INIT ();

xb_value_bindings_bind_str (xb_query_context_get_bindings (&context),
			        0,
			        term,
			        NULL);
n = xb_node_query_with_context (cpt_node, query, &context, NULL);
``` 

So, the stemmed `strategi` does not show up in the full text of `summary`, `"A strategy game"`, it has however been added to the node as token before. So I would expect this code to find the node and yield a result. This is not what happens though.

In XbSilo's `xb_silo_machine_func_search_cb`, only `op2` has `XB_OPCODE_FLAG_TOKENIZED` set, so we do not perform the fast TOKEN/TOKEN search: https://github.com/hughsie/libxmlb/blob/main/src/xb-silo.c#L1520

Instead, the code *always* falls back to the slower full-text search below, which in this case does not yield any result.

Why does `op1` (containing the "strategi" string) not have that one as token, and why does this inevitably always fall back to the slower search path? Am I supposed to bind the query value differently? Dirty hacks like just adding the string as a token as well to the opcode do not seem to work, so I think I am missing something here / do not understand how tokens relate to text search in libxmlb in concept.

I'm glad for any help :-)

hughsie:
xb_silo_machine_fixup_attr_search_token_cb() should be tokenizing the op1, no? If you could do a small extra [failing] test in ./src/xb-self-test.c I can take a look tomorrow.

ximion:
[...]

PR #144, opened 2023-08-14
DRAFT: Add example to reproduce #143

ximion:
Hi!

This just adds some test code to reproduce the issue described in #143.
It uses an unconventional, manual way to construct the query, which leads to the node containing the token not being found.

### DIFF
--- a/src/xb-self-test.c
+++ b/src/xb-self-test.c
@@ -1720,6 +1720,77 @@ xb_xpath_func(void)
 }
 
 static void
+xb_manual_token_search_func(void)
+{
+	XbNode *n;
+	g_autoptr(XbNode) cpt_node = NULL;
+	g_autoptr(XbBuilder) builder = NULL;
+	g_autoptr(XbBuilderNode) bn_root = NULL;
+	g_autoptr(XbBuilderNode) bn = NULL;
+	g_autoptr(XbSilo) silo = NULL;
+	g_autofree gchar *str = NULL;
+	g_autoptr(XbQuery) query = NULL;
+	g_autoptr(GPtrArray) result = NULL;
+	g_autoptr(GError) error = NULL;
+	g_auto(XbQueryContext) context = XB_QUERY_CONTEXT_INIT();
+
+	/* create node tree and add a dummy token manually */
+	bn_root = xb_builder_node_new("component");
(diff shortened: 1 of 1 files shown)


hughsie:
So, the issue seems to be:

            /* search for token (fails) */
    +       xb_silo_set_profile_flags(silo, ~0);
            query = xb_query_new(silo, "summary[text()~=?]", &error);

gives:

	# XbMachine-DEBUG: opcodes_sig=FUNC:text,?INT,FUNC:search
	# XbMachine-DEBUG: before optimizing: text(),?0,search()
	# XbMachine-DEBUG: ignoring optimized call to search(text(),?0): ?INT:FUNC types not supported
	# XbMachine-DEBUG: ignoring optimized call to text(): cannot optimize: no silo to query
	# XbMachine-DEBUG: after optimizing: text(),?0,search()
	# XbSilo-DEBUG: 0.01ms      query on component with `summary[text(),?0,search()]` [?0 → strategi] limit=0 -> 0 results

i.e. it's not matching `FUNC:text,TEXT,FUNC:search` as it's not a literal text value, it's a bound-at-query-time string. This means we can't optimize the query when it's created and we need to do it after calling `xb_value_bindings_bind_str()`.

Does wip/hughsie/query-bound-tokenize make sense and work for you? If it does, then I'll rebase that and squash it down to make sense and do a PR.

ximion:
> i.e. it's not matching `FUNC:text,TEXT,FUNC:search` as it's not a literal text value, it's a bound-at-query-time string. This means we can't optimize the query when it's created and we need to do it after calling `xb_value_bindings_bind_str()`.

Ah, this suddenly makes so much sense! I really confused myself when investigating the issue ^^ 

Your proposed patch makes a lot of sense to me, I will test it later today (when I am at the right computer...) and report back. I do wonder whether there is any way I can work around this issue in AppStream... I could not bind the string at query time, but then I'd need to carefully sanitize the value beforehand to ensure it doesn't cause issues... It would also be a lot less flexible.
If this works as expected though, the search improvement for AppStream should be so massive that depending on a more recent libxmlb is probably the correct solution.

hughsie:
Replaced by https://github.com/hughsie/libxmlb/pull/145

PR #145, opened 2023-08-15
Wip/hughsie/query bound tokenize

hughsie:


src/xb-self-test.c
```diff
@@ -1719,6 +1719,79 @@ xb_xpath_func(void)
...
+	g_assert_no_error(error);
+	g_assert_nonnull(str);
+	g_debug("\n%s", str);
+
+	/* get component node */
+	cpt_node = xb_silo_query_first(silo, "component", &error);
+	g_assert_no_error(error);
+	g_assert_nonnull(cpt_node);
+
+	/* search using the slow, token-less path (works) */
```
ximion, line 1763: We should probably drop the (works) from my original example ;-)

src/xb-self-test.c
```diff
@@ -1719,6 +1719,79 @@ xb_xpath_func(void)
...
+	g_assert_no_error(error);
+	g_assert_nonnull(result);
+
+	n = XB_NODE(g_ptr_array_index(result, 0));
+	g_assert_nonnull(n);
+	g_assert_cmpstr(xb_node_get_text(n), ==, "A strategy game");
+	g_clear_object(&query);
+	g_ptr_array_unref(g_steal_pointer(&result));
+
+	/* search for token (fails) */
```
ximion, line 1779: ...same for the (fails) - all of this should be working now.

ximion (approved):
Looks good to me and makes perfect sense - just maybe drop the works/fails from my original comment strings before merging.

Thank you for looking into this!

src/xb-self-test.c
hughsie: Done.

src/xb-self-test.c
hughsie: Fixed.

PR #146, opened 2023-08-17
trivial: Only tokenize when using ~= for the first time

hughsie:
We don't want the penalty of tokenizing for the most common == operator.


ximion (approved):
Just tested it, it works fine and the change makes sense to me as well!

hughsie:
@ximion release tagged

ximion:
> @ximion release tagged

Thank you! I'll make a new release of the AppStream stable series as well then :-)

PR #148, opened 2023-08-24
Fix comparing indexed text with integer values

hughsie:
Both are xb_opcode_cmp_val() but it only makes sense to compare `int:int` and `itx:itx` rather than `itx:int`.

### DIFF
--- a/src/xb-machine.c
+++ b/src/xb-machine.c
@@ -1475,7 +1475,7 @@ typedef gboolean (*OpcodeCheckFunc)(XbOpcode *op);
 static gboolean
 _xb_opcode_cmp_val_or_str(XbOpcode *op)
 {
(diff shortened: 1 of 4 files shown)


mcrha:
This is for https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2275 and I can confirm I see the `New & Updated` apps again with this patch applied on top of the main branch.

hughsie:
@mcrha merge and release? or should we wait for some review from @pwithnall ?

ximion (approved):
With my very recently gained understanding of this code, this change makes sense and shouldn't break AppStream (quite the contrary, I think AppStream may also be affected by this bug - I will have a look  later today).
So, looks good to me!

pwithnall:
Don’t wait for me! I’m unlikely to be able to review this for a few days

PR #151, opened 2023-09-17
trivial: Abort the compile when MSVC is detected

hughsie:


ximion (approved):

ximion:
That's a good idea! I should add something similar to my projects as well!
If MSVC ever does add support, this would need to become a versioned check of course - but I assume Microsoft would rather switch to Clang completely than to expend its C support...

PR #152, opened 2023-09-25
Windows: Support building with clang-cl

fanc999-1:
Hi,

This attempts to support building libxmlb for Windows using clang-cl, so that Visual Studio 2015 (or so, or later) is able to utilize libxmlb as a dependency, where the following is addressed:

*  Fix linking the libxmlb DLL, as shared code cannot be `inline`ed and be used across different compilation units in at least the clang-cl builds (let me know if having the `inline`ed implementation in the headers themselves is preferred).
*  Use `__declspec(dllexport)` to export symbols from the libxmlb DLL (static builds using clang-cl is not yet supported before or after this PR)-this is required for clang-cl to export the libxmlb symbols, sadly, and requires decorating public symbols as needed.
* Make `generate-version-script.py` output `libxmlb.map` with LR line endings, so that the `libxmlb-exported-api` test won't fail due to differing line endings (clang-cl builds using the Visual Studio command prompt, meaning Python will output in CR/LF line endings by default).

With blessings, thank you!

### DIFF
--- a/src/gen-def-from-map.py
+++ b/src/gen-def-from-map.py
@@ -1,0 +1,100 @@
+#!/usr/bin/python3
+# pylint: disable=invalid-name,missing-docstring
+#
+# Copyright (C) 2023 Chun-wei Fan <fanc999@yahoo.com.tw>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+# Script to generate .def file from .map files
+
+import os
+import re
+import sys
+
+def get_sym_groups(lines, namespace):
+    sym_groups = []
+    search_header = True
+    collect_symbols = False
+    search_tail = False
+
+    # namespace + version of library
+    namespace_with_ver_regex = re.escape(namespace) + \
+                           r"_[0-9]+\.[0-9]+\.[0-9]+"
+    # Look for symbols added in version in library from map file,
+    # which defines a group of symbols
+    header_regex = re.compile(r"^" + namespace_with_ver_regex)
+    # Look for where to start collecting symbols for the group
+    global_regex = re.compile(r"^\s+global:\s+$")
+    # Look for where to end collecting symbols for the group
+    local_regex = re.compile(r"^\s+local:\s+\*+;\s+$")
+    # Look, if any, since which version were the symbols
+    # group added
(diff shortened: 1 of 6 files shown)


src/xb-exports.h
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#pragma once
+
+/* clang-cl also defines _MSC_VER */
+#if defined (_MSC_VER) && defined (LIBXMLB_BUILD_DLL)
+# define LIBXMLB_API __declspec (dllexport)
```
hughsie, line 11: this isn't ideal -- is there any way to change the behaviour so that anything in the .map file is exported? Otherwise we have to "export" the function in two places.

fanc999-1: Hi,

Yeah, I understand what you mean.  Perhaps I will try to see whether there is a way for a script to pull out the `xb_*` symbols from the .map file to generate a .def file (needed for Visual Studio-like builds, if not going by the `__declspec(dllexport)` route, or even update the script that is used to generate the .map file to be able to generate the .def file that is needed.

fanc999-1: Hi,

I think this is now ready as I added a script to go through the `libxmlb.map` file to generate the .def file that is now needed and used to export the symbols from the libxmlb DLL as we are now not going through the `__decspec(dllexport)` route.

With blessings, thank you!

hughsie: I've merged 3/5 of those patches, thanks. For the .def file -- would it be better to generate it from the introspection data rather than the .map file? It would be a lot less complex code (which I'm happy to write!) but it would mean depending on introspection for the clan-cl build.

fanc999-1: Hi,

The dilemma is, we would need the symbols to be exported from the DLL before we could build the introspection files, so this would become a circular situation…

hughsie: Right, understood.

hughsie: I ran the python file through black and merged it -- thanks.

PR #154, opened 2023-10-17
xb-builder: Release file handles early

mcrha:
The XbBuilderSource::istream can be a file input stream holding a file handle to the underlying file. These can be released early, as soon as they are processed, thus they do not consume opened file handles needed for other operations.

This had been reported downstream:  
https://bugzilla.redhat.com/show_bug.cgi?id=2244575

### DIFF
--- a/src/xb-builder-source.c
+++ b/src/xb-builder-source.c
@@ -437,6 +437,7 @@ xb_builder_source_get_istream(XbBuilderSource *self, GCancellable *cancellable, 
 {
 	XbBuilderSourcePrivate *priv = GET_PRIVATE(self);
 	g_autofree gchar *basename = NULL;
+	g_autoptr(GInputStream) istream = NULL;
 	GFile *file;
 
 	g_return_val_if_fail(XB_IS_BUILDER_SOURCE(self), NULL);
@@ -446,8 +447,8 @@ xb_builder_source_get_istream(XbBuilderSource *self, GCancellable *cancellable, 
 		return g_object_ref(priv->istream);
 
 	/* convert the file to a GFileInputStream */
-	priv->istream = G_INPUT_STREAM(g_file_read(priv->file, cancellable, error));
-	if (priv->istream == NULL)
+	istream = G_INPUT_STREAM(g_file_read(priv->file, cancellable, error));
+	if (istream == NULL)
 		return NULL;
 
 	/* run the content type handlers until we get application/xml */
@@ -458,7 +459,7 @@ xb_builder_source_get_istream(XbBuilderSource *self, GCancellable *cancellable, 
 		XbBuilderSourceAdapter *item;
 		g_autofree gchar *content_type = NULL;
 		g_autoptr(GInputStream) istream_tmp = NULL;
-		g_autoptr(XbBuilderSourceCtx) ctx = xb_builder_source_ctx_new(file, priv->istream);
+		g_autoptr(XbBuilderSourceCtx) ctx = xb_builder_source_ctx_new(file, istream);
 
 		/* get the content type of the stream */
 		xb_builder_source_ctx_set_filename(ctx, basename);
@@ -486,7 +487,7 @@ xb_builder_source_get_istream(XbBuilderSource *self, GCancellable *cancellable, 
 		if (istream_tmp == NULL)
 			return NULL;
 		xb_builder_source_remove_last_extension(basename);
-		g_set_object(&priv->istream, istream_tmp);
+		g_set_object(&istream, istream_tmp);
 
 		/* the #GFile is only useful for the outermost input stream,
 		 * for example it points to the .tar.gz file, while inner input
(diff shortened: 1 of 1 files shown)


hughsie:
Hey! Back a bit; why are the input streams outliving the XbBuilder? I thought when the XbBuilder is destroyed, all the XbBuilderSource's would also be destroyed -- and at that point the input streams would be unref'd.

mcrha:
As a reproducer in gnome-software, delete `~/.cache/gnome-software/appstream/components.xmlb` and then run gnome-software and check opened file handles when on line https://gitlab.gnome.org/GNOME/gnome-software/-/blob/1afe9dc2215c74f1abbd654200fd6656fcd19f79/plugins/core/gs-plugin-appstream.c#L838 , aka after `xb_builder_ensure` call inside `gs_plugin_appstream_check_silo` and you'll see all the files from `/usr/share/metainfo` and other directories being opened. They are released together with the XbBuilder structure, but it is too late when they are thousands of the files to be processed (with 1024 file handles limit set in the system).

src/xb-builder.c
```diff
@@ -300,6 +300,10 @@ xb_builder_compile_source(XbBuilderCompileHelper *helper,
 
 	/* success */
 	xb_silo_add_profile(helper->silo, timer, "compile %s", guid);
+
+	/* free the istream, to release the file handle, because it's not needed anymore */
+	xb_builder_source_unset_istream(source);
```
hughsie, line 305: I do wonder if we could use g_ptr_array_copy() (with func==-NULL) and then remove it from the priv->sources GPtrArray instead...

src/xb-builder-source.c
```diff
@@ -500,6 +500,16 @@ xb_builder_source_get_istream(XbBuilderSource *self, GCancellable *cancellable,
 	return g_object_ref(priv->istream);
 }
 
+void
+xb_builder_source_unset_istream(XbBuilderSource *self)
+{
+	XbBuilderSourcePrivate *priv = GET_PRIVATE(self);
+	g_return_if_fail(XB_IS_BUILDER_SOURCE(self));
+	/* keep memory input streams, which cannot be recreated in xb_builder_source_get_istream() */
+	if (priv->istream != NULL && !G_IS_MEMORY_INPUT_STREAM(priv->istream))
```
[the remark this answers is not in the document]
mcrha: due to xb_builder_source_load_xml() and xb_builder_source_load_bytes()

src/xb-builder.c
mcrha: I'm not sure I follow. If you can call xb_builder_compile multiple times, then that would cause problems, no?

src/xb-builder-source.c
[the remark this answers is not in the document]
hughsie: Right, but we're done with accessing all input streams at this point -- no?

src/xb-builder.c
hughsie: Yes, but we can change the behaviour to disallow calling it multiple times with the opt-in XB_BUILDER_COMPILE_FLAG_INVALIDATE_SOURCES flag

mcrha:
I had on mind another solution, maybe cleaner, but I went with this one. The other one is to not cache file streams in the xb_builder_source_get_istream(). Then you either get the memory input stream or the file is read again.

By the way, the input streams are one-way, are they not? Unless they support GSeekable and they are rewind-ed to the beginning of the stream on the xb_builder_compile() call you do not get the same result when you call the xb_builder_compile() twice in a row, on the same sources.

mcrha:
I just pushed "the other solution", discarding original changes.

hughsie:
> you do not get the same result when you call the xb_builder_compile() twice in a row, on the same sources

Yes, I suppose that's true -- as in we've accidentality got the requirement that you can only call compile() one in reality. Maybe we could just cement that into reality and add it as a comment in the gtk-doc. I think changing xb_builder_source_get_istream() might change the semantics a bit -- so maybe just remove the source in all cases then?

src/xb-builder-source.c
```diff
@@ -438,6 +438,7 @@ xb_builder_source_get_istream(XbBuilderSource *self, GCancellable *cancellable,
 	XbBuilderSourcePrivate *priv = GET_PRIVATE(self);
 	g_autofree gchar *basename = NULL;
 	GFile *file;
+	GInputStream *istream;
```
hughsie, line 441: maybe do g_autoptr(GInputStream) istream = NULL; here and then return g_steal_pointer(&istream) -- otherwise you'll leak this on error.

hughsie:
> might change the semantics a bit

Maybe not -- your patch might be okay -- I thought you were going to rip out the GInputStream private instance.

src/xb-builder-source.c
mcrha: Ouch, you are right, it can leak otherwise. I'll update the merge request.

hughsie: If that passes all the tests then I think we've got something good -- would you need a new release/tarball?

mcrha: I just updated the merge request.

I'm not sure about a new release, it was the first time I saw it, thus nothing common. Feel free to wait few weeks, maybe more fixes will pile up. In any case, I'll left it up to you.

hughsie (approved):

PR #160, opened 2024-03-30
Hughsie/xz

hughsie:


ximion (approved):
This looks good!
While generally more compile-time options make it really hard to target a unified version of xmlb with a defined featureset, the fact that zstd is also optional makes making lzma optional only logical as well.

The changes also are very straightforward :-)

PR #170, opened 2024-04-09
Fix decompressing large zstd files, harder

hughsie:
This partially reverts 20eeebd4b385da1c7165cf3e9d94acd449d7a2bc, but instead moves the success check to after we update bytes_read and bytes_written.

Fixes https://github.com/fwupd/fwupd/issues/7060


superm1:
Yup, this fixes it, thanks!

Issue #189, opened 2024-06-27
Leading/trailing whitespace sometimes lost


mcrha:
Moving this from a downstream bug report:  
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2607

When checking the `<description>`, the text can have dropped leading/trailing whitespace, when it contains a new-line character (`\n`). The whitespace is left as is when the text does not contain any `\n`.

<details><summary>Reproducer is here.</summary>

```c
/* gcc `pkg-config --cflags --libs glib-2.0 xmlb` -Wno-deprecated-declarations xmlb.c -g -O0 -o xmlb && ./xmlb */

#include <glib.h>
#include <xmlb.h>
#include <locale.h>

static void
traverse_node (XbNode *node,
	       gint indent)
{
	g_autoptr(XbNode) sibling = NULL;
	g_autoptr(XbNode) next = NULL;
	const gchar *text;

	if (!node)
		return;

	g_print ("%*selem:'%s'", indent * 3, "", xb_node_get_element (node));

	text = xb_node_get_text (node);
	if (text)
		g_print (" text:'%s'", text);

	text = xb_node_get_tail (node);
	if (text)
		g_print (" tail:'%s'\n", text);
	else
		g_print ("\n");

	next = xb_node_get_child (node);
	if (next)
		traverse_node (next, indent + 1);
	g_clear_object (&next);

	for (sibling = xb_node_get_next (node); sibling != NULL; sibling = g_steal_pointer (&next)) {
		next = xb_node_get_next (sibling);
		traverse_node (sibling, indent);
		g_clear_object (&sibling);
	}
}

int main (int argc,
	  const gchar *argv[])
{
	g_autoptr(XbSilo) silo = NULL;
	g_autoptr(GError) error = NULL;
	g_autoptr(XbBuilder) builder = NULL;
	g_autoptr(XbBuilderSource) builder_source = NULL;
	const gchar *const *locales = g_get_language_names ();
	XbNode *node;

	setlocale (LC_ALL, "");

	builder_source = xb_builder_source_new ();
	if (!xb_builder_source_load_xml (builder_source,
		"<component>"
		  "<description>"
		    "<p>first paragraph</p>"
		    "<p>second <b>paragraph</b> with\nnewline <code>code</code> too.</p>"
		  "</description>"
		"</component>",
		XB_BUILDER_SOURCE_FLAG_NONE, &error)) {
		g_warning ("Failed to load XML: %s", error->message);
		return -1;
[...]

PR #190, opened 2024-06-28
Do not always strip literal text

hughsie:
Fixes https://github.com/hughsie/libxmlb/issues/189


hughsie:
@mcrha new package building in https://copr.fedorainfracloud.org/coprs/rhughes/fwupd/builds/

mcrha:
This is better , thank you. It also uncovers some changes are needed on the gnome-software side as well.

PR #209, opened 2024-10-14
Hughsie/fuzzing

hughsie:


hughsie:
@spencerwuwu your review please.

hughsie:
@superm1 I was tempted to release this today too. The fuzz fix is kinda important.

superm1 (approved):

PR #213, opened 2024-10-15
Hughsie/ancestor regression

hughsie:


hughsie:
I'm fuzzing this now, but so far so good.

sidt4:
Thanks for the quick fix Hughsie.

Should we have a unit test for this fix ?

hughsie:
> Should we have a unit test for this fix ?

That's probably a good idea. I think it's trying to handle a *lot* of deep children, so it's not a trivial thing to test.

sidt4:
Maybe `appstream` (or other `libxmlb` clients) has tests which can test `libxmlb` in these non-trivial cases.

Issue #236, opened 2025-03-05
xb_silo_query() returns NULL without error set


mcrha:
I have an interesting problem, I think it is a problem. When I run a certain query using `xb_silo_query` on one .xmlb, the function returns `NULL`, but not `GError`. I thought it's assured that the `NULL` is an error state, but maybe the things changed (or my expectation is wrong).

[This is a test program xmlb.c](https://github.com/user-attachments/files/19091565/xmlb.c.txt) (with added `.txt` extension to be able to attach it here) exhibiting the problem when I run it as:

```
$ ./xmlb ~/.cache/gnome-software/flatpak-user-user/components.xmlb
Exported xmlb as 50915 letters

** (process:34959): WARNING **: 16:17:41.568: Failed to execute query: no-error-returned
```

The .xz-compressed .xmlb file has 2.9MB, which feels like a lot for here, but I can share it by other means. I tried to regenerate it too (within gnome-software), but it did not cure the problem.

hughsie:
Can you email it to richard_at_hughsie.com pls?

hughsie:
I can reproduce here, perhaps easier with:

    $ ./src/xb-tool query components.xmlb "components/component/keywords/keyword[text()~=stem('calen')]"
    Segmentation fault (core dumped)

hughsie:
So, using a *different* xmlb file, I get:

	./src/xb-tool query ~/.cache/gnome-software/appstream/components.xmlb "components/component/keywords/keyword[text()~=stem('calen')]"
	RESULT: <keyword>Calendar</keyword>
	RESULT: <keyword>calendar</keyword>
	RESULT: <keyword>Calendar</keyword>
	RESULT: <keyword>calendar</keyword>
	RESULT: <keyword>Calendar</keyword>

...so it's something specific to the components.xmlb you sent me. _Iiiiiinnnnnteresting_.

PR #237, opened 2025-03-05
Always return a GError when a query fails

hughsie:
If the silo is damaged (perhaps disk corruption) then we might get unset when querying for the node text.

The `xb_silo_from_strtab()` function already checks for this (setting a GError) so we do not need to do the early check -- which was wrong anyway.

Additionally, get the node text (which can fail) *before* pushing the new opcode so that valgrind is happy that XbOpcode is uninit.

Fixes https://github.com/hughsie/libxmlb/issues/236


mcrha:
> If the silo is damaged (perhaps disk corruption) then we might get unset when querying for the node text.

Is it damaged? I can reproduce even when building the .xmlb file from scratch, thus if broken, it is built broken, not made broken afterwards.

mcrha:
Here is [an updated test app](https://github.com/user-attachments/files/19104246/xmlb.c.txt), it only reads whichever appstream.xml(.gz) file(s) you pass to it and runs the test. I sent you a link for one appstream.xml.xz (better compression, needs to decompress), with which I can reproduce with the libxmlb `main` branch at commit e59d8f6f108ebcc1cf09f46476dafa138708ca41 . It's a file from the flathub.org.

Notices:
* the "components.xmlb" is constructed always from the appstream data, and even it is saved to the disk, it's not read from it, thus one can rule out the disk corruption
* the exported XML is significantly smaller (only one component) and shows a corruption here (the `<releases/>` at the end of the file, completely out of its expected position)
* it does not matter whether I use `xb_builder_compile` or `xb_builder_ensure`, neither using `0` for the flags for these functions has any impact

hughsie:
Okay, good hunting! It looks like this fix was just papering over the issue and not actually solving the root cause. Minimizing your XML file a bit, this seems to be the problem:

input:

    <?xml version="1.0" encoding="UTF-8"?>
    <components version="0.8">
      <component>
        <content_rating type="xxx"/>
        <releases>
          <release version="yyy"/>
        </releases>
      </component>
    </components>

output:

    <components version="0.8">
      <component>
        <content_rating type="xxx" />
      </component>
      <releases>
        <release version="yyy" />
      </releases>
    </components>

hughsie:
> this seems to be the problem

So that was an export problem -- a bug fixed in 9790a5b056a60bd62e637df4bc681d056c56786d -- but not the one I'm looking for.

hughsie:
Looking at the xmlb dump, and exporting that section I see:

    <keywords>
      ...
      <keyword>Yandex Translate</keyword>
      <keyword>
      </keyword>
    </keywords>

an empty keyword!

hughsie:
@mcrha okay, worked it out! I've pushed this new branch and am building an rpm in the COPR. If that works for you I'll merge and make a new tag.

mcrha:
No need for the COPR, I build from the source. Let me give it a try...

mcrha:
It seems to work almost the same as the `0.2.1`. Diffing the output of it and the latest `main` with this change on top of it I see corrections of the empty `<release>` tags (and some other, which used to be done in two lines before) and also possibly unrelated:

```diff
-      <p>Both checkboxes need to be ticked for the plugin to function!</p>
+      <p>    Both checkboxes need to be ticked for the plugin to function!     </p>
```

which comes from

```xml
      <p>
    Both checkboxes need to be ticked for the plugin to function!
    </p>
```
I guess it's intentional to preserve the trailing spaces (I think it was asked, maybe by me, long time ago, but I can be wrong).

mcrha:
> Yes, intentional. We could add a flag to turn it off if you'd like.

Not needed, at the moment. Thanks.

PR #245, opened 2025-04-26
Fix heap churn by arena allocating temporary nodes during compile stage

### LINKED ISSUE #244: Reduce heap churn when parsing XML
Derived from https://gitlab.gnome.org/GNOME/gnome-software/-/issues/941

libxmlb parses an xml with a two-step process. First it reads the data and turns it into a temporary in-memory tree of XML nodes, then it traverses the tree and serializes it to the compact representation.

The temporary nodes are all heap allocated, so this introduces a lot of heap churn. A profile of gnome-software showed XML parsing with libxmlb was the primary source of short-lived heap allocations triggering heap fragmentation.

The lifetime of these builder nodes is tightly scoped to the `xb_builder_compile` function call. Arena allocating builder nodes should improve performance and reduce impact to other users of the heap.

omc8db:
(WIP, it's my first time contributing so this is going to need some work. I'm sending this in a rough state to get eyes on it early.)

Resolves #244 

Moves all temporary allocations created by compiling an XML into a memory arena - primarily builder nodes and their children.

Sanity check: Compile the Fedora software catalog XML to XMLB with this tool before and after changes. Verify output is the same.

```
owen@owen-thinkpad:~/repos/libxmlb$ /usr/bin/time -v xb-tool compile before.xmlb fedora.xml.gz 
        Command being timed: "xb-tool compile before.xmlb fedora.xml.gz"
        User time (seconds): 0.89
        System time (seconds): 0.07
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.97
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 172976
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 41764
        Voluntary context switches: 7
        Involuntary context switches: 13
        Swaps: 0
        File system inputs: 0
        File system outputs: 21464
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
owen@owen-thinkpad:~/repos/libxmlb$ /usr/bin/time -v build/src/xb-tool compile after.xmlb fedora.xml.gz 
        Command being timed: "build/src/xb-tool compile after.xmlb fedora.xml.gz"
        User time (seconds): 0.44
        System time (seconds): 0.06
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.51
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 135392
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 32421
        Voluntary context switches: 9
        Involuntary context switches: 6
        Swaps: 0
        File system inputs: 0
        File system outputs: 21496
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
owen@owen-thinkpad:~/repos/libxmlb$ xb-tool export before.xmlb > /tmp/before.xml
owen@owen-thinkpad:~/repos/libxmlb$ xb-tool export after.xmlb > /tmp/after.xml
owen@owen-thinkpad:~/repos/libxmlb$ diff /tmp/before.xml /tmp/after.xml 
owen@owen-thinkpad:~/repos/libxmlb$
```

Summary: Identical output, 20% less peak RAM usage and 50% faster execution time. 

Heaptrack results on this parsing benchmark (github won't let me upload the raw profiles here)
* Before changes: 3.29 million heap allocations
* After changes: 0.04 million heap allocations

More important than total RAM usage, these are now the type of allocation that glibc's malloc implementation can easily release back to the OS when compilation is complete. So this should really improve allocator pressure and heap fragmentation when libxmlb compile functions are called in a long-lived program.

Still needs work:
* This necessarily changes the public API around builder nodes. You can't just heap allocate them as GObjects now. A search showed code that uses this part of the API is very rare. The only thing I could find that depended on it was one function in the gnome-software appstream plugin. But it's still an API break and a rollout concern.
* I'm getting warnings about .gir file generation related to the public API change. Could use some help fixing that.
* (Cleanup) unit tests, formatting

### DIFF
--- a/src/xb-arena.c
+++ b/src/xb-arena.c
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2025 Owen Chiaventone <ochiaventone@gmail.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ */
+
+#include "xb-arena.h"
+
+#include <glib.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+
+struct XbArena {
+	size_t n_chunks;
+	size_t chunks_cap;
+	/* Array of pointers to chunks */
+	uint8_t **chunks;
+	/* Points to first free byte in current chunk */
+	uint8_t *tail;
+	uint8_t *current_chunk_end;
+};
+
+XbArena *
+xb_arena_new(void)
+{
+	XbArena *arena = malloc(sizeof(XbArena));
+	*arena = (XbArena){
+	    .n_chunks = 0,
+	    .chunks_cap = 8,
+	    .chunks = (uint8_t **)malloc(8 * sizeof(uint8_t *)),
+	    .tail = NULL,
+	    .current_chunk_end = NULL,
+	};
+	return arena;
+}
+
+static void
+xb_arena_add_chunk(XbArena *arena, uint8_t *chunk)
+{
+	/* Grow chunklist if needed */
+	if (arena->n_chunks + 1 >= arena->chunks_cap) {
+		void *new_chunklist = malloc(arena->chunks_cap * sizeof(uint8_t *) * 2);
+		memcpy(new_chunklist, (void *)arena->chunks, arena->n_chunks * sizeof(uint8_t *));
+		free((void *)arena->chunks);
+		arena->chunks = (uint8_t **)new_chunklist;
+		arena->chunks_cap *= 2;
+	}
+	arena->chunks[arena->n_chunks] = chunk;
+	arena->n_chunks++;
+}
+
+void *
+xb_arena_alloc(XbArena *arena, size_t len)
+{
+	void *alloc = NULL;
+	if (len == 0)
+		return NULL;
+
+	/* Treat allocations larger than a chunk size as their own chunk
+	 * It's not worth trying to optimize those. If this case hits a lot, try
+	 * increasing the arena chunk size. */
+	if (len > XB_ARENA_CHUNKSIZE) {
+		uint8_t *chunk = malloc(len);
+		xb_arena_add_chunk(arena, chunk);
+		/* No need to invalidate tail or current_chunk_end, there's still space
+		in the current chunk for more allocations */
+		return chunk;
+	}
+	/* Round allocations up to nearest whole machine word
+	 * If we didn't do this, allocated structs may cost extra
+	 * to access because of unaligned load/stores and may
+	 * straddle multiple cache lines */
+	len = (len + sizeof(ssize_t) - 1) & ~(sizeof(ssize_t) - 1);
+
+	/* Allocate another chunk if the the requested size won't fit in the current one.
+	 * This will always be hit for the first allocation on a new arena
+	 * because tail and current_chunk_end are initialized to zero. */
+	if (arena->tail + len > arena->current_chunk_end) {
+		uint8_t *chunk = malloc(XB_ARENA_CHUNKSIZE);
+		xb_arena_add_chunk(arena, chunk);
+		arena->tail = chunk;
+		arena->current_chunk_end = chunk + (XB_ARENA_CHUNKSIZE);
+	}
+
+	alloc = arena->tail;
+	arena->tail += len;
+	return alloc;
+}
+
+void
+xb_arena_free(XbArena *arena)
+{
+	for (uint32_t i = 0; i < arena->n_chunks; i++) {
+		free(arena->chunks[i]);
+	}
+	free((void *)arena->chunks);
+	free(arena);
+}
+
+char *
+xb_arena_strdup(XbArena *arena, const char *src)
+{
+	size_t n = 0;
+	char *out = NULL;
+	if (!src)
+		return NULL;
+	n = strlen(src) + 1;
+	out = xb_arena_alloc(arena, n);
+	memcpy(out, src, n);
+	out[n] = '\0';
+	return out;
+}
+
+char *
+xb_arena_strndup(XbArena *arena, const char *src, size_t strsz)
+{
+	size_t n = 0;
+	char *out = NULL;
+	if (!src)
+		return NULL;
+	/* strnlen isn't in the c99/posix standard even though it's widely supported
+	 * easy enough to do it ourselves */
+	for (n = 0; n < strsz; n++) {
+		if (src[n] == '\0')
+			break;
+	}
+	out = xb_arena_alloc(arena, n + 1);
+	memcpy(out, src, n);
+	out[n] = '\0';
+	return out;
+}
+
+XbArenaPtrArray *
+xb_arena_ptr_array_new(XbArena *arena)
+{
+	XbArenaPtrArray *self = xb_arena_alloc(arena, sizeof(XbArenaPtrArray));
+	*self = (XbArenaPtrArray){
+	    .arena = arena,
+	    .pointers = (void **)xb_arena_alloc(arena, 4 * sizeof(void *)),
+	    .len = 0,
+	    .cap = 4,
+	};
+	return self;
+}
+void
+xb_arena_ptr_array_add(XbArenaPtrArray *self, void *data)
+{
+	/* Grow array if necessary */
+	if (self->len + 1 > self->cap) {
+		void *newptrs = NULL;
+		self->cap *= 2;
+		newptrs = xb_arena_alloc(self->arena, self->cap * sizeof(void *));
+		memcpy(newptrs, (void *)self->pointers, self->len * sizeof(void *));
+		self->pointers = (void **)newptrs;
+	}
+	/* Insert new pointer */
+	self->pointers[self->len] = data;
+	self->len++;
+}
+void
+xb_arena_ptr_array_remove(XbArenaPtrArray *self, void *data)
+{
+	for (guint i = 0; i < self->len; i++) {
+		if (self->pointers[i] == data) {
+			xb_arena_ptr_array_remove_index(self, i);
+			return;
+		}
+	}
+}
+
+void
+xb_arena_ptr_array_remove_index(XbArenaPtrArray *self, size_t index)
+{
+	/* Must preserve order of elements. Slide down all elements after the removed one */
+    for (guint i = index; i < self->len - 1; i++) {
+        self->pointers[i] = self->pointers[i+1];
+    }
+    self->len--;
+}
--- a/src/xb-arena.h
+++ b/src/xb-arena.h
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2025 Owen Chiaventone <ochiaventone@gmail.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ */
+
+#pragma once
+
+/* Simple arena allocator used to construct nodes during parsing pass
+ * If the nodes and strings on them were individually heap allocated instead, we would churn the
+ * heap with a large number of small and medium sized allocations. This is bad for performance in
+ * a short lived program, but more importantly can trigger serious heap fragmentation if you parse
+ * a large XML in the middle of a long-lived program.
+ * This was implicated in https://gitlab.gnome.org/GNOME/gnome-software/-/issues/941
+ *
+ * We don't know how large the XML is when we start parsing, so the arena may have to grow.
+ * This arena allocator uses a chunking strategy - if we run out, allocate another large block.
+ *
+ * This arena is *not* threadsafe. It is expected that it will be created and used from a single
+ * thread. Making the arena threadsafe would signficantly change its performance characteristics
+ */
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+/* default to 1MB chunks for arena allocation.
+ * Most allocators have heuristics to optimize large allocations
+ * For example with glibc malloc this should be above DEFAULT_MMAP_THRESHOLD
+ * see  https://sourceware.org/glibc/wiki/MallocInternals */
+#define XB_ARENA_CHUNKSIZE 1 << 20
+
+typedef struct XbArena XbArena;
+
+XbArena *
+xb_arena_new(void);
+
+/* Allocate bytes from the arena */
+void *
+xb_arena_alloc(XbArena *arena, size_t len);
+
+/* Free the entire arena at once. Deactivates the arena object. */
+void
+xb_arena_free(XbArena *arena);
+
+/*
+ * Arena friendly alternatives for GLib data structures
+ */
+char *
+xb_arena_strdup(XbArena *arena, const char *src);
+char *
+xb_arena_strndup(XbArena *arena, const char *src, size_t strsz);
+
+typedef struct XbArenaPtrArray {
+	XbArena *arena;
+	void **pointers;
+	size_t len;
+	size_t cap;
+} XbArenaPtrArray;
+
+XbArenaPtrArray *
+xb_arena_ptr_array_new(XbArena *arena);
+void
+xb_arena_ptr_array_add(XbArenaPtrArray *self, void *data);
+void
+xb_arena_ptr_array_remove(XbArenaPtrArray *self, void *data);
+void
+xb_arena_ptr_array_remove_index(XbArenaPtrArray *self, size_t index);
--- a/src/xb-builder-node.c
+++ b/src/xb-builder-node.c
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: LGPL-2.1-or-later
  */
 
-#include "glib.h"
 #define G_LOG_DOMAIN "XbSilo"
 
 #include "config.h"
@@ -18,6 +17,7 @@
 #include "xb-string-private.h"
 
 struct XbBuilderNode {
+	XbArena *arena; /* Memory arena to make any temporary allocations into */
 	guint32 offset;
 	gint priority;
 	XbBuilderNodeFlags flags;
@@ -27,22 +27,22 @@ struct XbBuilderNode {
 	guint32 text_idx;
 	gchar *tail;
 	guint32 tail_idx;
-	XbBuilderNode *parent; /* noref */
+	struct XbBuilderNode *parent; /* noref */
 
 	/* Around 87% of all XML nodes have zero children, so this array is only
 	 * allocated if it’s non-empty. %NULL means an empty array. */
-	GPtrArray *children; /* (element-type XbBuilderNode) (nullable) */
+	XbArenaPtrArray *children; /* (element-type XbBuilderNode) (nullable) */
 
 	/* Around 80% of all XML nodes have zero attributes, so this array is only
 	 * allocated if it’s non-empty. %NULL means an empty array. */
-	GPtrArray *attrs; /* (element-type XbBuilderNodeAttr) (nullable) */
+	XbArenaPtrArray *attrs; /* (element-type XbBuilderNodeAttr) (nullable) */
 
 	/* Most nodes will have no tokens */
+	/* Todo: Arena destructor does not iterate over elements so these destructors
+	 * are never called. Switch these to arena allocated types or they will leak. */
 	GPtrArray *tokens;  /* (element-type utf8) (nullable) */
 	GArray *token_idxs; /* (element-type guint32) (nullable) */
 };
-static void
-xb_builder_node_attr_free(XbBuilderNodeAttr *attr);
 
 /**
  * xb_builder_node_has_flag:
@@ -82,7 +82,7 @@ xb_builder_node_add_flag(XbBuilderNode *self, XbBuilderNodeFlags flag)
 
 	self->flags |= flag;
 	for (guint i = 0; self->children != NULL && i < self->children->len; i++) {
-		XbBuilderNode *c = g_ptr_array_index(self->children, i);
+		XbBuilderNode *c = self->children->pointers[i];
 		xb_builder_node_add_flag(c, flag);
 	}
 }
@@ -115,8 +115,9 @@ xb_builder_node_get_element(XbBuilderNode *self)
 void
 xb_builder_node_set_element(XbBuilderNode *self, const gchar *element)
 {
-	g_free(self->element);
-	self->element = g_strdup(element);
+	/* We deliberately do not free the old element when setting a new one, since
+	 * freeing is managed by the arena. */
+	self->element = xb_arena_strdup(self->arena, element);
 }
 
 /**
@@ -139,7 +140,7 @@ xb_builder_node_get_attr(XbBuilderNode *self, const gchar *name)
 		return NULL;
 
 	for (guint i = 0; i < self->attrs->len; i++) {
-		XbBuilderNodeAttr *a = g_ptr_array_index(self->attrs, i);
+		XbBuilderNodeAttr *a = self->attrs->pointers[i];
 		if (g_strcmp0(a->name, name) == 0)
 			return a->value;
 	}
@@ -223,7 +224,7 @@ xb_builder_node_get_tail(XbBuilderNode *self)
 
 /* private */
 /* Returns NULL if the array is empty */
-GPtrArray *
+XbArenaPtrArray *
 xb_builder_node_get_attrs(XbBuilderNode *self)
 {
 	return self->attrs;
@@ -236,6 +237,7 @@ xb_builder_node_parse_literal_text(XbBuilderNode *self, const gchar *text, gssiz
 	guint newline_count = 0;
 	g_auto(GStrv) split = NULL;
 	gsize text_len_safe;
+	char *out = NULL;
 
 	/* sanity check */
 	if (text == NULL)
@@ -244,7 +246,7 @@ xb_builder_node_parse_literal_text(XbBuilderNode *self, const gchar *text, gssiz
 	/* we know this has been pre-fixed */
 	text_len_safe = text_len >= 0 ? (gsize)text_len : strlen(text);
 	if (xb_builder_node_has_flag(self, XB_BUILDER_NODE_FLAG_LITERAL_TEXT))
-		return g_strndup(text, text_len_safe);
+		return xb_arena_strndup(self->arena, text, text_len_safe);
 
 	/* all whitespace? */
 	if (xb_string_isspace(text, text_len_safe))
@@ -252,7 +254,7 @@ xb_builder_node_parse_literal_text(XbBuilderNode *self, const gchar *text, gssiz
 
 	/* all on one line, no trailing or leading whitespace */
 	if (g_strstr_len(text, text_len, "\n") == NULL)
-		return g_strndup(text, text_len_safe);
+		return xb_arena_strndup(self->arena, text, text_len_safe);
 
 	/* split the text into lines */
 	tmp = g_string_sized_new((gsize)text_len_safe + 1);
@@ -283,8 +285,10 @@ xb_builder_node_parse_literal_text(XbBuilderNode *self, const gchar *text, gssiz
 		newline_count = 1;
 	}
 
-	/* success */
-	return g_string_free(tmp, FALSE);
+	/* success. Return a copy of the sanizited string owned by the node's memory arena */
+	out = xb_arena_strdup(self->arena, tmp->str);
+	g_string_free(tmp, TRUE);
+	return out;
 }
 
 /**
@@ -411,26 +415,24 @@ xb_builder_node_set_attr(XbBuilderNode *self, const gchar *name, const gchar *va
 	g_return_if_fail(name != NULL);
 
 	if (self->attrs == NULL)
-		self->attrs =
-		    g_ptr_array_new_with_free_func((GDestroyNotify)xb_builder_node_attr_free);
+		self->attrs = xb_arena_ptr_array_new(self->arena);
 
 	/* check for existing name */
 	for (guint i = 0; i < self->attrs->len; i++) {
-		a = g_ptr_array_index(self->attrs, i);
+		a = self->attrs->pointers[i];
 		if (g_strcmp0(a->name, name) == 0) {
-			g_free(a->value);
-			a->value = g_strdup(value);
+			a->value = xb_arena_strdup(self->arena, value);
 			return;
 		}
 	}
 
 	/* create new */
-	a = g_slice_new0(XbBuilderNodeAttr);
-	a->name = g_strdup(name);
+	a = xb_arena_alloc(self->arena, sizeof(XbBuilderNodeAttr));
+	a->name = xb_arena_strdup(self->arena, name);
 	a->name_idx = XB_SILO_UNSET;
-	a->value = g_strdup(value);
+	a->value = xb_arena_strdup(self->arena, value);
 	a->value_idx = XB_SILO_UNSET;
-	g_ptr_array_add(self->attrs, a);
+	xb_arena_ptr_array_add(self->attrs, a);
 }
 
 /**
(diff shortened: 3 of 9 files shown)


hughsie:
I don't think we've done any profiling on memory use when building (unlike loading and querying), there are probably quite a bit of low hanging fruit. There's some easy wins in https://github.com/hughsie/libxmlb/pull/246 -- and probably a lot more. I suspect making the XbBuilderNode internal struct smaller is probably what we want to do.

omc8db:
Reducing the size of the XbBuilderNode internal struct is definitely worth looking into, but unfortunately I wasn't able to find any way to make XbBuilderNodes stay as GObjects and also be manually arena allocated. 

I tried something like #246 first but it still resulted in GObject making individual heap allocations under the hood for every builder node. And if I manually put the GObjects in an arena anyway by skipping g_object_new, GObject still made internal heap allocations for every builder node that got leaked when the arena was freed.

Maybe someone better at GObject than I am could figure out a way to make this work. If there's a way to fully manually allocate all per-object memory used by it including internal allocations, you could probably make this change without a public API break. Or at least a smaller one.

meson.build
omc8db, line 1: Oops, this was supposed to be a temporary local change but it slipped into the commit. But I am a little confused by the original warning list because Wdeclaration-after-statement conflicts with c_std=c99. 

Looks like the warning list is close to GCC's old default -Wall set for C89. I'm OK to undo changes to this file if you think it's too off topic, but at some point should probably ensure all the warning flags are valid for std=c99

src/xb-arena.c
```diff
@@ -0,0 +1,137 @@
...
+        xb_arena_add_chunk(arena, chunk);
+        /* No need to invalidate tail or current_chunk_end, there's still space
+        in the current chunk for more allocations */
+        return chunk;
+    }
+    /* Round allocations up to nearest whole machine word
+     * If we didn't do this, allocated structs may cost extra
+     * to access because of unaligned load/stores and may
+     * straddle multiple cache lines */
+    len = (len+sizeof(ssize_t)-1) & ~(sizeof(ssize_t)-1);
```
pwithnall, line 49: If available, I think you should align to the alignment of [`max_align_t`](https://en.cppreference.com/w/c/types/max_align_t), which aiui should give you the same fundamental alignment guarantees as `malloc()` itself.

src/xb-arena.c
```diff
@@ -0,0 +1,137 @@
...
+
+    /* Allocate another chunk if the the requested size won't fit in the current one.
+     * This will always be hit for the first allocation on a new arena
+     * because tail and current_chunk_end are initialized to zero. */
+    if(arena->tail + len > arena->current_chunk_end) {        
+        uint8_t* chunk = malloc(XB_ARENA_CHUNKSIZE);
```
pwithnall, line 55: Thinking about this, is there a particular reason for using `malloc()` to allocate the chunks? I think one of the advantages of using an arena allocator is that we can avoid libc’s heap management entirely.

So instead of allocating the chunks using `malloc()`, you could allocate them with `mmap(MAP_ANONYMOUS | MAP_PRIVATE)` to request an entire page from the OS (so the chunk size would become equal to the OS page size). This would limit you to allocations which are at most the size of an OS page (unless you want to introduce bookkeeping to allow an allocation to cross page boundaries), but that’s probably not an unreasonable limitation to impose on the caller, given that this is a special purpose allocator? Do you expect any huge allocations? If the input file needs to be allocated in memory it should probably be `mmap()`ed anyway.

Using `mmap()` for the chunks would remove a whole load of overhead from libc, and within the arena allocator, and reduce a number of potential failure paths too.

src/xb-arena.c
pwithnall: However, you may want to consider doing the alignment at the _head_ of an allocation (i.e. setting `alloc = arena->tail + alignment_offset`) rather than by incrementing the `len` of the previous allocation, as this will allow for tighter packing of small allocations: the alignment requirement is limited by the requested `len` of an allocation, so small allocations don’t need to be 16-byte aligned (for example).

src/xb-arena.c
```diff
@@ -0,0 +1,137 @@
...
+        .pointers = (void**) xb_arena_alloc(arena, 4 * sizeof(void*)),
+        .len = 0,
+        .cap = 4,
+    };
+    return self;
+}
+void xb_arena_ptr_array_add(XbArenaPtrArray* self, void* data) {
+    /* Grow array if necessary */
+    if(self->len+1 > self->cap) {
+        self->cap *= 2;
```
pwithnall, line 113: Various potential arithmetic overflows here need to be checked for

src/xb-arena.c
```diff
@@ -0,0 +1,137 @@
...
+    return out;
+}
+
+XbArenaPtrArray* xb_arena_ptr_array_new(XbArena* arena) {
+    XbArenaPtrArray* self = xb_arena_alloc(arena, sizeof(XbArenaPtrArray));
+    *self = (XbArenaPtrArray){
+        .arena = arena,
+        .pointers = (void**) xb_arena_alloc(arena, 4 * sizeof(void*)),
+        .len = 0,
+        .cap = 4,
```
pwithnall, line 106: Was this cap chosen based on some testing? It would be quite a penalty to reallocate the ptr array within the arena if it needs to grow (because we can guarantee that a reallocation is going to require a `memcpy`, whereas in a normal libc heap it sometimes doesn’t), so might be good to always slightly oversize the ptr array for an average usecase. Or allow the caller to specify the expected array length, so the arena code doesn’t have to guess.

src/xb-arena.h
```diff
@@ -0,0 +1,60 @@
...
+/* Simple arena allocator used to construct nodes during parsing pass
+ * If the nodes and strings on them were individually heap allocated instead, we would churn the
+ * heap with a large number of small and medium sized allocations. This is bad for performance in
+ * a short lived program, but more importantly can trigger serious heap fragmentation if you parse
+ * a large XML in the middle of a long-lived program.
+ * This was implicated in https://gitlab.gnome.org/GNOME/gnome-software/-/issues/941
+ *
+ * We don't know how large the XML is when we start parsing, so the arena may have to grow.
+ * This arena allocator uses a chunking strategy - if we run out, allocate another large block. 
+ */
```
pwithnall, line 18: Typically the approach in GLib-ish projects is to put doc comments like this in a gtk-doc block near the top of the C file for the utility (so, `xb-arena.c` in this case, below the `#include`s), as that’s where people look for them and that’s the file which is more likely to be read and modified. We’ve found this gives a greater chance of the comment being kept up to date with future modifications.

src/xb-arena.h
```diff
@@ -0,0 +1,60 @@
...
+    /* Array of pointers to chunks */
+    uint8_t** chunks;
+    /* Points to first free byte in current chunk */
+    uint8_t* tail;
+    uint8_t* current_chunk_end;
+} XbArena;
+
+XbArena xb_arena_new(void);
+
+// Allocate bytes from the arena
```
pwithnall, line 38: Similarly these comments would all move to being gtk-doc comments immediately above the function implementations in the C file.

Grep the rest of the source tree for `/**` for examples.

src/xb-arena.c
```diff
@@ -0,0 +1,137 @@
...
+        void** newptrs = xb_arena_alloc(self->arena, self->cap * sizeof(void*));
+        memcpy(newptrs, self->pointers, self->len * sizeof(void*));
+        self->pointers = newptrs;
+    }
+    /* Insert new pointer */
+    self->pointers[self->len] = data;
+    self->len++;
+}
+void xb_arena_ptr_array_remove(XbArenaPtrArray* self, void* data) {
+    for (guint i = 0; i < self->len; i++) {
```
pwithnall, line 123: (Putting this comment here arbitrarily, but it applies to the whole file)

What’s the thread safety situation for the arena allocator? Can we guarantee that it (and its utilities, like `xb_arena_ptr_array_remove()`) will only ever be called from a single thread for a given `XbArena`?

src/xb-builder-node.h
```diff
@@ -51,15 +38,50 @@ typedef enum {
 	XB_BUILDER_NODE_FLAG_LAST
 } XbBuilderNodeFlags;
 
+// XbBuilderNodes are deliberately not part of the GObject system. They are plain C structs
+// this is because there are potentially tens of thousands of them and the overhead of GObject
+// per-object RAM overhead and memory management becomes significant at that scale.
+// In versions 0.3.22 and earler these were GObjects, and profiling showed this was the main
+// bottleneck to parser performance.
```
pwithnall, line 45: This comment should probably become part of a gtk-doc comment for `XbBuilderNode`, in the C file.

src/xb-builder-node.h
```diff
@@ -51,15 +38,50 @@ typedef enum {
...
+typedef struct XbBuilderNode {
```
pwithnall, line 46: Does the `XbBuilderNode` struct have to move into the header file? Typically we want to keep structs opaque, as that enforces encapsulation of them, which makes code more maintainable.

Changing the way this one is allocated shouldn’t change that, I think.

src/xb-builder.c
```diff
@@ -657,7 +656,6 @@ xb_builder_import_node(XbBuilder *self, XbBuilderNode *bn)
 {
 	XbBuilderPrivate *priv = GET_PRIVATE(self);
 	g_return_if_fail(XB_IS_BUILDER(self));
-	g_return_if_fail(XB_IS_BUILDER_NODE(bn));
```
pwithnall, line 660: You could keep this precondition check in, but change it to `bn != NULL` (one of the implications of `XB_IS_BUILDER_NODE(bn)` is that `bn` must be non-`NULL`)

src/xb-self-test.c
```diff
@@ -9,6 +9,7 @@
 #include <gio/gio.h>
 #include <locale.h>
 
+#include "xb-arena.h"
```
pwithnall, line 12: Looking at the existing unit tests made me think it might be a good idea to have some explicit unit tests for `XbArena`, just to check corner case behaviour which would cause hard-to-debug problems if it went wrong.

For example:
 - Zero-sized allocations always return `NULL`
 - Returned allocated pointers are always appropriately aligned for the requested size
 - Corner cases with the utility functions (like `strdup` of `NULL` or of an empty string)

src/xb-builder-node.c
```diff
@@ -79,19 +47,17 @@ xb_builder_node_has_flag(XbBuilderNode *self, XbBuilderNodeFlags flag)
 void
 xb_builder_node_add_flag(XbBuilderNode *self, XbBuilderNodeFlags flag)
 {
-	XbBuilderNodePrivate *priv = GET_PRIVATE(self);
-	g_return_if_fail(XB_IS_BUILDER_NODE(self));
 
-	if ((priv->flags & flag) != 0)
+	if ((self->flags & flag) != 0)
```
pwithnall, line 51: All these changes from `priv` → `self` are good (and necessary), but a pain to review. In the final version of this PR, I strongly suggest they should be in a separate, preparatory commit, so they’re easier to review separately from any allocation/logic changes to `xb-builder-node.c` (or other files).

That commit could actually be PR #246, with this one based on top. Even though #246 keeps `XbBuilderNode` as a GObject, it does all the preparatory `priv` → `self` changes we need.

src/xb-builder-node.c
```diff
@@ -265,17 +220,19 @@ xb_builder_node_parse_literal_text(XbBuilderNode *self, const gchar *text, gssiz
...
 	if (xb_string_isspace(text, text_len_safe))
 		return NULL;
 
 	/* all on one line, no trailing or leading whitespace */
 	if (g_strstr_len(text, text_len, "\n") == NULL)
-		return g_strndup(text, text_len_safe);
+		return xb_arena_strndup(self->arena, text, text_len_safe);
 
 	/* split the text into lines */
+	// Todo: This makes temporary heap allocations, but it's not on the hot path.
```
pwithnall, line 234: Note: TODO comment here needs addressing (just highlighting it so it doesn’t get forgotten about, I realise this is still work in progress)

pwithnall (commented):
> Summary: Identical output, 20% less peak RAM usage and 50% faster execution time.
> 
> Heaptrack results on this parsing benchmark (github won't let me upload the raw profiles here)
> 
>     Before changes: 3.29 million heap allocations
>     After changes: 0.04 million heap allocations

These stats are absolutely fantastic :heart:

I would say the API changes are unavoidable, and worth the performance improvements. The final decision for that rests with Richard though, I’m merely a drive-by reviewer with libxmlb. From gnome-software’s point of view, we’d be happy to conditionally compile against new and old versions of the libxmlb API to get these improvements.

It isn’t possible to keep `XbBuilderNode` as a GObject and allocate it through a custom allocator, GObject explicitly doesn’t support that. It’s just too complex.

I’ve left a load of comments below, most of them are discussion points or random thoughts rather than concrete things to change. I think there are a couple of bigger questions which would benefit from being answered first, like:
 * Could the arena allocator use `mmap()` rather than `malloc()` to request pages as chunks?
 * What’s the story about thread safety?

I didn’t fully read through all the changes in `xb-builder-node.c` because my eyes started glazing over, so that’ll need doing before this actually lands.

Since you’ve run heaptrack on it, I guess that means you’ve also run valgrind on it. It would be useful to also build it with asan to make sure asan is happy about things — I could imagine that an arena allocator might make asan unhappy because it can’t track allocations as well. Even if we’re not leaking memory or doing undefined behaviour, I think we still need to keep asan happy, or we lose the ability to use that as a debugging tool elsewhere in the process in future.

src/xb-arena.c
```diff
@@ -0,0 +1,137 @@
...
+    return (XbArena){
+        .n_chunks = 0,
+        .chunks_cap = 8,
+        .chunks = malloc(8 * sizeof(uint8_t*)),
+        .tail = NULL,
+        .current_chunk_end = NULL,
+    };
+}
+
+static void xb_arena_add_chunk(XbArena* arena, uint8_t* chunk) {
```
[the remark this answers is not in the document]
omc8db: Yep, definitely need to fix this up. Just didn't get to it yet for the proof of concept.

meson.build
omc8db: I don't have a super strong opinion and you're right it's better to match the style of the existing code - fair enough. I'll undo the change to this file and line up the rest of the code.

omc8db:
> * Could the arena allocator use `mmap()` rather than `malloc()` to request pages as chunks?

Definitely possible, but glibc already has this  under the hood for large allocations: [source](https://sourceware.org/glibc/wiki/MallocInternals#Malloc_Algorithm:~:text=If%20the%20request%20is%20large%20enough%2C%20mmap()%20is%20used%20to%20request%20memory%20directly%20from%20the%20operating%20system.)

I'll at least add a comment with that info.

That's what I meant when I said this is the type of allocation that glibc's malloc implementation can easily release back to the OS - free on one of these large chunks immediately unmaps instead of doing fancy accounting to try to keep it in one of the small allocation pools that stays around forever.

Other C mallocs I've seen (jemalloc, tcmalloc) do similar things. Haven't checked MUSL yet, I'll look into that and make sure it does what we expect here too.

> * What’s the story about thread safety?

The story is xb_compile is single threaded :)

One of the reasons an arena is so much faster here is that there's no data structures shared with other threads or mutexes to lock when doing a malloc.

You did get me thinking about what happens if code outside the library calls xb_builder_node_new or xb_builder_node_insert. If that's not on the same thread trouble might happen. But I don't think those functions were threadsafe before this change either - xb_builder_node_insert writes a field in *parent without locking or atomics. To be honest I'm not sure why these internal details were exposed in the public API in the first place.

Because this is on a proven hot path, there's an unfortunate cost to adding thread safety here that needs to be carefully considered.

> Since you’ve run heaptrack on it, I guess that means you’ve also run valgrind on it. It would be useful to also build it with asan to make sure asan is happy about things — I could imagine that an arena allocator might make asan unhappy because it can’t track allocations as well. Even if we’re not leaking memory or doing undefined behaviour, I think we still need to keep asan happy, or we lose the ability to use that as a debugging tool elsewhere in the process in future.

yep! ASAN is fine with arena allocators as long as you don't access out of the bounds of your arena. It's one of the reasons for using libc malloc for chunks instead of rolling my own with mmap - asan gives better diagnostics this way and it was helpful while I was getting the proof of concept going. ASAN will not be able to tell if you access spills between object boundaries in the arena, but that's the unavoidable nature of them.

The tricky thing to keep in mind now is we can't have any GLib managed objects owned by a builder node as free functions are never run on individual objects in the arena.

src/xb-builder-node.c
omc8db: Sounds good. I'll rearrange the commits on my next push

pwithnall:
> > * What’s the story about thread safety?
> 
> The story is xb_compile is single threaded :)
> 
> One of the reasons an arena is so much faster here is that there's no data structures shared with other threads or mutexes to lock when doing a malloc.
> 
> You did get me thinking about what happens if code outside the library calls xb_builder_node_new or xb_builder_node_insert. If that's not on the same thread trouble might happen. But I don't think those functions were threadsafe before this change either - xb_builder_node_insert writes a field in *parent without locking or atomics. To be honest I'm not sure why these internal details were exposed in the public API in the first place.

Sounds good. Perhaps I should have been clearer, sorry: I was asking to make sure the thread safety requirements of the arena were known (and documented). I wasn’t trying to suggest that it should be made threadsafe, or that I believe the APIs are used across threads in the wild (they shouldn’t be, aiui).

I think if we’re explicit in the documentation for `XbArena` and `XbBuilder` that they must be used single-threaded, that would cover everything that’s needed from a thread safety point of view.

src/xb-builder-node.h
omc8db: It doesn't need to move. Updated to keep struct opaque

src/xb-builder-node.c
omc8db: Changes split out into commit 6403b1a0bd90e448ed0ff55120ea6bff28c51a44

omc8db:
@pwithnall  the latest push works and passes the selftests but builds with these warnings about generating GObject Interospection bindings. What's the best way to resolve them? I don't see an easy way to satisfy the automatic suggestions without breaking arena allocation.

```
[19/21] Generating src/Xmlb-2.0.gir with a custom command (wrapped by meson to set env)
../src/xb-builder-node.c:584: Warning: Xmlb: Unknown container Type(ctype=XbArenaPtrArray*) for element-type annotation
../src/xb-builder-node.c:584: Warning: Xmlb: xb_builder_node_get_children: return value: Unresolved type: 'XbArenaPtrArray*'
../src/xb-builder-node.c:643: Warning: Xmlb: xb_builder_node_get_child: return value: Invalid non-constant return of bare structure or union; register as boxed type, add (copy-func) and (free-func), or (skip)
../src/xb-builder-node.c:549: Warning: Xmlb: xb_builder_node_get_parent: return value: Invalid non-constant return of bare structure or union; register as boxed type, add (copy-func) and (free-func), or (skip)
../src/xb-builder-node.c:827: Warning: Xmlb: xb_builder_node_new: argument arena: Unresolved type: 'XbArena*'
../src/xb-builder-node.c:833: Warning: Xmlb: xb_builder_node_new: return value: Invalid non-constant return of bare structure or union; register as boxed type, add (copy-func) and (free-func), or (skip)
../src/xb-builder-node.c:827: Warning: Xmlb: xb_builder_node_new: argument arena: Unresolved type: 'XbArena*'
../src/xb-builder-node.c:833: Warning: Xmlb: xb_builder_node_new: return value: Invalid non-constant return of bare structure or union; register as boxed type, add (copy-func) and (free-func), or (skip)
```

hughsie:
I'm really confused why a custom arena would have such a big performance increase compared to just using glib-wrapped-malloc -- surely the glibc and kernel code has been optimized to crazy over the years? Is the thread-safety really 20%?

pwithnall:
> I'm really confused why a custom arena would have such a big performance increase compared to just using glib-wrapped-malloc -- surely the glibc and kernel code has been optimized to crazy over the years? Is the thread-safety really 20%?

My assumption is that the biggest win comes from reducing the number of heap allocations (calls to `malloc()`) by chunking them (i.e. `xb_arena_malloc()` only calls libc’s `malloc()` if it’s run out of space). That effectively turns most memory allocations in `XbBuilder` into a pointer increment and dereference. That’s also what gives us the heap usage wins, by forcing all the arena allocations into one heap area which can be freed without fragmentation once the `XbBuilder` is done.

There will also be a win from not using GObject, as it does a couple of allocations for each constructed GObject, vs the one needed by using a plain struct.

The `malloc()` thread safety probably won’t contribute that much overall, because libc’s malloc uses a per-thread arena anyway. But it certainly doesn’t hurt to avoid the overhead of the TLS lookup to find the arena on each `malloc()` call.

omc8db:
> I'm really confused why a custom arena would have such a big performance increase compared to just using glib-wrapped-malloc -- surely the glibc and kernel code has been optimized to crazy over the years? Is the thread-safety really 20%?

Philip is correct. glibc malloc is well optimized, but it has the design constraint that you could call free() on any allocation at any time. That requires it to know where every allocation begins and ends, and where gaps are in memory that could be reused. The techniques malloc uses for this are clever and efficient but there's no free lunch. An arena doesn't let you free individual elements, so it doesn't have this per-allocation memory overhead. That's where the 20% lower RAM usage is coming from.

The doubled CPU performance mainly comes from three places
* At the end you no longer have to visit every node in the tree individually to free it. This was invisibly handled by GObject cleanup functions before, but it cost nearly as much real CPU time as traversing the entire tree again.
* xb_arena_malloc does less work per call than malloc because it's just a pointer bump vs [the glibc malloc algorithm](https://sourceware.org/glibc/wiki/MallocInternals#:~:text=thread%20Cache%20(tcache)-,Malloc%20Algorithm,-In%20a%20nutshell)
* Traversing the tree accesses data in roughly the same order it was added. malloc tries to group data together by allocation size, so data added at the same time may go to different CPU cache lines. The arena puts data added at the same time together regardless of size, so it has a better CPU cache hit rate for this workload.

Multithreading support is not the biggest factor in performance, but in practice this part of the code was not multithreaded. Worrying about it would mean more code complexity and maybe slightly worse performance for questionable gain.

src/xb-builder.c
```diff
@@ -803,17 +802,20 @@ xb_builder_compile(XbBuilder *self,
 		XbBuilderSource *source = g_ptr_array_index(priv->sources, i);
 		const gchar *prefix = xb_builder_source_get_prefix(source);
 		g_autofree gchar *source_guid = xb_builder_source_get_guid(source);
-		g_autoptr(XbBuilderNode) root = NULL;
+		XbBuilderNode *root = NULL;
```
hughsie, line 805: This kind of API break is a huge pain for people using the library. I really think we should avoid changing how people are using the public libxmlb API.

hughsie:
I'm not totally onboard with these changes; at least from a Red Hat point of view I'd have to explain exactly why exactly I'm using a different allocater and make it work with the various valgrind and sanitizer tests we have for RHEL. I can imagine it would be a huge amount of "paperwork" to explain why various things are changing. It would be a pretty huge timesync to ensure that everything is still working with the old and new APIs as changed now.

The bit I'm struggling with is that libxmlb was always designed to be super quick and efficient to _query_ the silo, but the actual _generation_ of the silo CPU and memory use was almost a non-goal -- if we wanted to optimize both then we would have chosen a quite different silo internal layout to avoid having the string table as a huge blob at the end and take the hit of larger integer sizes as offsets. There would also have been better savings with regards to using GRefString everywhere rather than strndup'ing lots of small strings in various places.

I think my main thought is that big API changes like this also come with a maintenance cost; I we can make something that we do once per day 0.5s quicker then I don't think it's worth the soname bump or the added complexity. If we can make queries (which we do hundreds of thousands of every day) faster, even by a few ms then I do think that would be worthwhile.

That said, if you manage to keep the API compatible I'd be open to adding the valgrind and sanitizer bits.

pwithnall:
This isn’t so much about speed but about heap fragmentation. If gnome-software rebuilds its libxmlb caches once a day (due to updated repository metadata), and runs in the background, then that’s once a day that a large amount of heap fragmentation is caused. This increases gnome-software’s used heap page count (memory footprint) in a way which can’t be tidied up.

Memory usage is one of the top two issues people have with gnome-software, so this is not an unimportant thing.

I think it’s important enough that I’m happy to help out with any porting work to a new API, in consumers of libxmlb.

hughsie:
> then that’s once a day that a large amount of heap fragmentation is caused

Maybe it would be simpler to run a subprocess? You could even use xb-tool rather than make a new tool.

pwithnall:
> Maybe it would be simpler to run a subprocess? You could even use xb-tool rather than make a new tool.

Progress reporting and error handling from a subprocess is a nightmare.

src/xb-arena.c
```diff
@@ -0,0 +1,182 @@
...
+
+#include <glib.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+
+struct XbArena {
+	size_t n_chunks;
```
hughsie, line 17: needs to be gsize

src/xb-arena.c
```diff
@@ -0,0 +1,182 @@
...
+	uint8_t **chunks;
+	/* Points to first free byte in current chunk */
+	uint8_t *tail;
+	uint8_t *current_chunk_end;
+};
+
+XbArena *
+xb_arena_new(void)
+{
+	XbArena *arena = malloc(sizeof(XbArena));
```
hughsie, line 29: g_malloc

src/xb-arena.c
```diff
@@ -0,0 +1,182 @@
...
+		free((void *)arena->chunks);
+		arena->chunks = (uint8_t **)new_chunklist;
+		arena->chunks_cap *= 2;
+	}
+	arena->chunks[arena->n_chunks] = chunk;
+	arena->n_chunks++;
+}
+
+void *
+xb_arena_alloc(XbArena *arena, size_t len)
```
hughsie, line 56: gsize etc

src/xb-arena.c
```diff
@@ -0,0 +1,182 @@
...
+	arena->tail += len;
+	return alloc;
+}
+
+void
+xb_arena_free(XbArena *arena)
+{
+	for (uint32_t i = 0; i < arena->n_chunks; i++) {
+		free(arena->chunks[i]);
+	}
```
hughsie, line 98: no need for braces

src/xb-arena.c
```diff
@@ -0,0 +1,182 @@
...
+		return NULL;
+	n = strlen(src) + 1;
+	out = xb_arena_alloc(arena, n);
+	memcpy(out, src, n);
+	out[n] = '\0';
+	return out;
+}
+
+char *
+xb_arena_strndup(XbArena *arena, const char *src, size_t strsz)
```
hughsie, line 118: gsize and gchar etc

src/xb-arena.c
```diff
@@ -0,0 +1,182 @@
...
+	XbArenaPtrArray *self = xb_arena_alloc(arena, sizeof(XbArenaPtrArray));
+	*self = (XbArenaPtrArray){
+	    .arena = arena,
+	    .pointers = (void **)xb_arena_alloc(arena, 4 * sizeof(void *)),
+	    .len = 0,
+	    .cap = 4,
+	};
+	return self;
+}
+void
```
hughsie, line 148: needs newline between functions

src/xb-arena.h
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * We don't know how large the XML is when we start parsing, so the arena may have to grow.
+ * This arena allocator uses a chunking strategy - if we run out, allocate another large block.
+ *
+ * This arena is *not* threadsafe. It is expected that it will be created and used from a single
+ * thread. Making the arena threadsafe would signficantly change its performance characteristics
+ */
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
```
hughsie, line 24: just glib.h I think

src/xb-arena.h
```diff
@@ -0,0 +1,67 @@
...
+} XbArenaPtrArray;
+
+XbArenaPtrArray *
+xb_arena_ptr_array_new(XbArena *arena);
+void
+xb_arena_ptr_array_add(XbArenaPtrArray *self, void *data);
+void
+xb_arena_ptr_array_remove(XbArenaPtrArray *self, void *data);
+void
+xb_arena_ptr_array_remove_index(XbArenaPtrArray *self, size_t index);
```
hughsie, line 67: needs trailing newline

src/xb-builder-node.c
```diff
@@ -997,11 +835,18 @@ xb_builder_node_class_init(XbBuilderNodeClass *klass)
...
-	priv->element = g_strdup(element);
+	XbBuilderNode *self = xb_arena_alloc(arena, sizeof(XbBuilderNode));
+	*self = (XbBuilderNode){
+	    .arena = arena,
+	    .element = xb_arena_strdup(arena, element),
+	    .element_idx = XB_SILO_UNSET,
+	    .text_idx = XB_SILO_UNSET,
+	    .tail_idx = XB_SILO_UNSET,
+	    .attrs = NULL,    /* only allocated when an attribute is added */
+	    .children = NULL, /* only allocated when a child is added */
```
hughsie, line 848: i think we can assume it's all set to NULL/0 if we use g_malloc0

src/xb-builder-node.h
```diff
@@ -121,15 +114,10 @@ xb_builder_node_traverse(XbBuilderNode *self,
...
 gchar *
 xb_builder_node_export(XbBuilderNode *self, XbNodeExportFlags flags, GError **error)
     G_GNUC_NON_NULL(1);
 GPtrArray *
 xb_builder_node_get_tokens(XbBuilderNode *self) G_GNUC_NON_NULL(1);
 void
-xb_builder_node_add_token(XbBuilderNode *self, const gchar *token) G_GNUC_NON_NULL(1, 2);
-
-G_END_DECLS
+xb_builder_node_add_token(XbBuilderNode *self, const gchar *token) G_GNUC_NON_NULL(1, 2);
```
hughsie, line 123: needs newline at end

src/xb-self-test.c
```diff
@@ -2327,31 +2333,32 @@ xb_builder_node_token_max_func(void)
 static void
 xb_builder_node_func(void)
 {
+	XbArena *arena = xb_arena_new();
```
hughsie, line 2336: needs to be g_autoptr

src/xb-builder.c
```diff
@@ -788,11 +784,14 @@ xb_builder_compile(XbBuilder *self,
...
-	helper->root = xb_builder_node_new(NULL);
+	helper->arena = xb_arena_new();
+	helper->root = xb_builder_node_new(helper->arena, NULL);
 	helper->silo = xb_silo_new();
 	helper->locales = priv->locales;
 	helper->strtab = g_byte_array_new();
-	helper->strtab_hash = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+	/* Keys in this hash table are owned by the arena allocator, so the key destroy function is
+	 * NULL */
+	helper->strtab_hash = g_hash_table_new_full(g_str_hash, g_str_equal, NULL, NULL);
```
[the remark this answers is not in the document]
pwithnall: imo `g_hash_table_new_full()` is a bit clearer here because it explicitly shows that the free functions are set to `NULL`, which ties in with the comment above. But I’m really nitpicking on the shade of the garden shed here!

hughsie:
I'm going to close this. I really feel like this is a half-finished patch and it needs a lot of work.

PR #246, opened 2025-04-26
Hughsie/xb builder final

hughsie:


hughsie:
@pwithnall this would be an API break, correct? I can't remember if derivable -> final is okay.

pwithnall:
Correct, derivable → final is an API break. (Final → derivable is not an API break)

PR #310, opened 2026-05-18
Fix use-after-free in xb_machine_func_in_cb haystack building

hughsie:
When in() pops opcodes that own their strings (e.g. from lower-case()), the per-iteration g_auto(XbOpcode) frees the string at the end of each loop iteration while the haystack array still references it.

## Summary by CodeRabbit

* **Tests**
  * Added tests for XPath in() with transformed text and missing-needle cases.

* **Bug Fixes**
  * in() now returns "not found" for absent/null needles instead of erroring.

* **Refactor**
  * Streamlined in() predicate processing for reduced allocations and clearer logic; removed legacy fallback code paths.

* **Chores**
  * Added a new query-lookup API and deprecated the old one (now fails gracefully with a warning).
  * Updated GLib/build version constraints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/hughsie/libxmlb/pull/310?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -129,8 +129,8 @@ else
 endif
 mandir = join_paths(prefix, get_option('mandir'))
 
-gio = dependency('gio-2.0', version : '>= 2.45.8')
-giounix = dependency('gio-unix-2.0', version : '>= 2.45.8', required: false)
+gio = dependency('gio-2.0', version : '>= 2.68.0')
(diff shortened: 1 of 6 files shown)


hughsie:
@coderabbitai review

hughsie:
@coderabbitai review

PR #333, opened 2026-07-27
Hughsie/stale indexes

hughsie:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data reload handling by clearing stale cached information before reinitialising.
  * Ensures refreshed data is used consistently after a reload.


hughsie:
@coderabbitai review

PR #335, opened 2026-07-31
Fix a potential out of bounds read when building a XbBuilderNode

hughsie:
Although xb_builder_node_set_text() is normally NUL-terminated, it doesn't have to be. Split the string taking into account the possibility that NUL is missing.

Found by Aisle Research, many thanks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of length-limited text input, including buffers that are not NUL-terminated.
  * Preserved embedded newline content when processing builder-node text.

* **Tests**
  * Added coverage for zero-length input, NUL-terminated strings and embedded newline data.

### DIFF
--- a/src/xb-builder-node.c
+++ b/src/xb-builder-node.c
@@ -12,6 +12,7 @@
 #include <string.h>
 
 #include "xb-builder-node-private.h"
+#include "xb-common-private.h"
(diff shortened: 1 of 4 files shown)


hughsie:
@coderabbitai review

